PR #4, opened 2016-06-22
Naive initial rwlock implementation

JohnPJenkins:
Implementation is of the classic mutex+condition variable variety. Unoptimized (doesn't need to broadcast every time).

Test currently just checks for write safety in a multi-reader, multi-writer scenario. Am open to ideas for more robust testing, as in doing hacks to make sure that multiple readers are concurrently in the same critical region, or that readers and writers aren't in the region at the same time.


smseo:
First of all, thanks for the implementation. I think your rwlock implementation works correctly. A couple of comments.
1) The following function signatures need to be changed.
`int ABT_rwlock_rdlock(ABT_rwlock *rwlock);`
`int ABT_rwlock_wrlock(ABT_rwlock *rwlock);`
`int ABT_rwlock_unlock(ABT_rwlock *rwlock);`
-->
`int ABT_rwlock_rdlock(ABT_rwlock rwlock);`
`int ABT_rwlock_wrlock(ABT_rwlock rwlock);`
`int ABT_rwlock_unlock(ABT_rwlock rwlock);`

2) For more robust testing, first I would suggest adding iterations in the test code like `test/basic/barrier.c`. Then, to make sure concurrent accesses to the same critical region among readers, you can put a barrier (among readers) in the critical section of readers. They should be able to reach the barrier together. Finally, for the mutual exclusiveness between readers and writers, you may let only one reader (e.g., a specific id of reader) increase `g_counter` in the critical section then validate at the end. What do you think?

smseo:
That's a good question. We have handled deadlocks/hangs manually, but it would be good to have a systematic way to handle them. I'm not sure whether `automake` supports timeout. I need to look into it. For now, you can leave it without wrapping the test program in a timeout script.

smseo:
Thanks @JohnPJenkins. Your commits were merged to the master.

PR #20, opened 2017-06-15
Change the host detection mechanism in configure.ac to support POWER8.

### LINKED ISSUE #18: Power8 build fails due to ABT_CONFIG_USE_FCONTEXT being unset
Same error as in issue #11:

(make produces)

 CC       libabt_la-barrier.lo
In file included from ../src/include/abtd.h:48:0,
                 from ../src/include/abti.h:128,
                 from barrier.c:6:
../src/include/abtd_thread.h: In function 'ABTD_thread_context_set_arg':
../src/include/abtd_thread.h:162:2: error: #error "Not implemented yet"
 #error "Not implemented yet"

My configure command line is:
./configure CC=gcc
but same error occurs if default (xlc) compiler is used.

shintaro-iwasaki:
It resolves #18 by fixing `configure.ac`.


halimamer:
This patch is related to issue #18 but does not actually solve it. It enables `fcontext` support for Power8 arch, but the real issue in #18 is that `ucontext` support is not fully supported. You could say it fixes one of two problems.

PR #22, opened 2017-08-15
initialize unit_def thread field in thread_htable

carns:
- needed for traversal of queue (linked list) when there is more than
  one element present

This pull request fixes the bug described in https://lists.argobots.org/pipermail/discuss/2017-August/000014.html

### DIFF
--- a/src/thread_htable.c
+++ b/src/thread_htable.c
@@ -75,6 +75,7 @@ void ABTI_thread_htable_push(ABTI_thread_htable *p_htable, int idx,
         p_queue->head = p_thread;
         p_queue->tail = p_thread;
     } else {
+        p_thread->unit_def.thread = ABTI_thread_get_handle(p_thread);
         p_queue->tail->unit_def.p_next = &p_thread->unit_def;
         p_queue->tail = p_thread;
     }
@@ -101,6 +102,7 @@ ABT_bool ABTI_thread_htable_add(ABTI_thread_htable *p_htable, int idx,
         /* Change the ULT's state to BLOCKED */
         ABTI_thread_set_blocked(p_thread);
 
+        p_thread->unit_def.thread = ABTI_thread_get_handle(p_thread);
         p_queue->tail->unit_def.p_next = &p_thread->unit_def;
         p_queue->tail = p_thread;
     }
@@ -134,6 +136,7 @@ void ABTI_thread_htable_push_low(ABTI_thread_htable *p_htable, int idx,
         p_queue->low_head = p_thread;
         p_queue->low_tail = p_thread;
     } else {
+        p_thread->unit_def.thread = ABTI_thread_get_handle(p_thread);
         p_queue->low_tail->unit_def.p_next = &p_thread->unit_def;
         p_queue->low_tail = p_thread;
     }
@@ -160,6 +163,7 @@ ABT_bool ABTI_thread_htable_add_low(ABTI_thread_htable *p_htable, int idx,
         /* Change the ULT's state to BLOCKED */
         ABTI_thread_set_blocked(p_thread);
 
+        p_thread->unit_def.thread = ABTI_thread_get_handle(p_thread);
         p_queue->low_tail->unit_def.p_next = &p_thread->unit_def;
         p_queue->low_tail = p_thread;
     }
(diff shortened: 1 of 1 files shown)


halimamer:
@carns , what kind of pool are you using? If it's the FIFO one, then that field should have already been initialized at creation time. If you have a custom pool, then it might be that your `u_create_from_thread` function definition does not initialize that field. More specifically, the line responsible for this initialization is at https://github.com/pmodels/argobots/blob/master/src/thread.c#L78.

I have noticed that the other synchronization constructs also assume that field might not be initialized. I don't know the reason behind this assumption. In other words, I don't see why a thread calling a synchronization function would be created without initializing that field.

carns:
Good catch, @halimamer !

The custom scheduler that I am using requires a custom pool to go with it, so I am not using the default pool implementation in Argobots.  I copied the custom pool directly from the stock one to use as a starting point, though.  Here is what my unit_create_from_thread() function looks like:

https://xgitlab.cels.anl.gov/sds/abt-snoozer/blob/master/src/pool.c#L389

I copied that from the Argobots fifo pool at the time:

https://github.com/pmodels/argobots/blob/3550e10891deeaea402b0563043a06e5736ad296/src/pool/fifo.c#L421

.... but now the Argobots fifo pool has this implementation of the same function:

https://github.com/pmodels/argobots/blob/master/src/pool/fifo.c#L401

So!  I can probably solve this by updating my unit_create_from_thread() function to mimic what the default one was doing, in particular using the embedded unit def.  I will try that and report back.  It looks like I was lucky it didn't break worse.  Or maybe unlucky that it didn't break more obviously!

There is an underlying meta problem, though, which is that I created my custom pool and scheduler by copying the default implementations and modifying them to suite my needs, but then the responsibilities of the pool and/or scheduler changed without me realizing it.

In the short term I can fix this specific bug in my pool.  In the medium term I can make a complete pass over my pool and scheduler and update them to match the stock Argobots code in case something else slipped through.

In the longer term, any suggestions on how I can keep from getting out of sync?

halimamer:
I think the core problem comes from you allocating another instance of `unit_def` for the same thread. The storage of`unit_def` is part of the storage of the thread itself already allocated at https://github.com/pmodels/argobots/blob/master/src/thread.c#L52. Consequently, there exists two instances of `unit_def`, which caused this mess.

Part of the confusion comes from the function name `unit_create_from_thread`, which shouldn't have called `create` in the first place since it is not creating anything. It is simply an initialization function. There are many instances of such misleading naming that need to be fixed. I will create an issue in order to not forget about it.

To answer your question, I don't think you were out of sync; Argobots just needs better naming and documentation :)

halimamer:
I have started this new issue to track misleading naming in Argobots: https://github.com/pmodels/argobots/issues/24

carns:
Ok.  It looks like (sorry, trying to reverse engineer why the abt-snoozer code is set up why it is; I don't totally remember) I defined my own unit_t not because I needed something different in it, but because the pool implementation needed to access fields within that struct.  

The default unit definition is an ABTI type that isn't accessible to code compiled outside of the argobots repo.  As long as that's true I can't safely reuse the built-in unit struct field in the thread.

thread_htable.c directly accesses the built-in unit_def in several places, though, so its not really safe to define and alloc my own either :)

I could hack something up that (essentially) makes a copy of the struct ABTI_unit definition in my pool implementation so that I can use the unit_def field directly, but that's going to be very fragile.

Suggestions?

carns:
Actually, even if I could access the default unit struct type, that by itself wouldn't be enough because I can't get to the unit_def field without the ABTI_thread type as well.

It seems like there is a fundamental issue of the pool API allowing  custom pools to define/allocate their own unit structs vs. some pieces of Argobots assuming the pool is using the embedded unit struct.

I'm still looking at this and thinking about it, just sharing comments as I go.

halimamer:
Those are good points. I am also thinking about it.

carns:
Stepping back a little bit, I only need a pretty trivial change to the stock pool: I need to add a field to the pool's data_t struct, and I need to call one extra function (with that field as an argument) when something a new unit is pushed to a shared pool.

When I think about it that way, I wonder if I could create a pool that just "piggy-backs" on top of the default pool. Each function just calls the corresponding functions in the default fifo pool, its just that it intercepts the  data struct and wraps something else around it, and after calling the push function it executes the extra trigger on a field in that wrapper struct.

It will be a little bit before I can try hands-on to organize the code that way so I might be missing some problem that makes this impossible for some reason, but that would be ideal if I could stack them like this.  Not only would it dodge the internal structure problem but it would make sure that I'm always up to date with the default Argobots algorithm for the pool.

halimamer:
In general, I think the `unit_create_from_thread` does not need to be exposed to the user. If I remove this function from the definition, would that work for you without changing  your current code (apart deleting the code related to the above function naturally)?

halimamer:
My previous suggestion was bad. The right fix is to avoid overloading the `ABT_unit` type and to solve the user-defined vs. internally defined unit type and abstract the way it is accessed.

PR #27, opened 2017-11-11
eventual: added a nonblocking test API

halimamer:
A thread might want to overlap waiting on an eventual with other activities. The
current eventual interface only allow a blocking wait operation. This patch
extends the eventual interface with a nonblocking routine; ABT_eventual_test.


carns:
Looks good to me.  FYI @mdorier, I cherry picked this to https://github.com/carns/argobots/tree/dev-get-dev-basic as well if you want to work/test from there.

Issue #19, opened 2017-06-13
Compile-time cache line size detection


shintaro-iwasaki:
Now the cache line size is specified by setting the environmental variable (`ABT_CACHE_LINE_SIZE` or `ABT_ENV_CACHE_LINE_SIZE`). It imposes the runtime calculation of `gp_ABTI_global->mem_sh_size` and lengthens the critical path of ULT creation as a result (see `ABTI_mem_alloc_thread()` in `abti_mem.h`).

Cache line size should be detected and embedded into the binary at compile time.

shintaro-iwasaki:
#33 was merged.

PR #33, opened 2018-01-29
#19 Cache line size can be statically obtained on configuration.

shintaro-iwasaki:
Solve #19  It improves performance by embedding the constant.

Specifically, users can specify one of the following:
- auto: use the automatic cache line size detection. It uses 64 if all attempts fail.
- specific value (e.g., “64”): use that value.
- otherwise (default): use an environmental variable, or use 64 if it is not set.

Internally, the following command is used for Linux:
``cat /sys/devices/system/cpu/cpu0/cache/index0/coherency_line_size | grep -Po '^(\d+)$'``
I checked it on the following environments:
- Red Hat 7.4 + POWER8 ... OK
- Red Hat 7.4 + Intel Xeon ... OK
- Red Hat 7.4 + Intel Xeon Phi (KNL) ... OK
- Ubuntu 16.04 + Intel Xeon ... OK
- VMWare Ubuntu 16.04 + Intel Core i7 ... OK
- openSUSE 42.2 Arm64 (ARM Cortex 8) ... NG
Note I could not find any command working for the last environment.

For Mac, the following is used.
``sysctl -a | grep hw.cachelinesize | grep -Po '(\d+)$' | grep -Po '^(\d+)$'``
It worked on Intel Mac.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -193,6 +193,13 @@ AC_ARG_ENABLE([sched-sleep],
 AC_ARG_ENABLE([simple-mutex],
     AS_HELP_STRING([--enable-simple-mutex], [use a simple mutex implementation]))
 
+# --enable-static-cacheline-size
+AC_ARG_ENABLE([static-cacheline-size],
+[  --enable-static-cacheline-size@<:@=OPTS@:>@ embed cache line size at compile-time.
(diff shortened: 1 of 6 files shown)


halimamer:
test:jenkins

shintaro-iwasaki:
test:jenkins

Issue #34, opened 2018-01-29
Test fails on my laptop.


shintaro-iwasaki:
The test sometimes fails when I execute it on my laptop.
```
$ make check
...
PASS: sched_user_ws
PASS: pool_access
PASS: mutex
PASS: mutex_prio
FAIL: mutex_recursive
...
```
Specifically, it shows the following error nondeterministically:
```
$ ./mutex_recursive
lt-mutex_recursive: mutex.c:656: ABTI_mutex_wait: Assertion `rank < p_htable->num_rows' failed.
Aborted
```

It seems because my laptop has few cores.
The following seems happening.
1. `p_global->num_cores` is set to the number of cores.
https://github.com/pmodels/argobots/blob/master/src/arch/abtd_env.c#L29
2.  `p_global->max_xstreams` is (by default) set to `p_global->num_cores`.
https://github.com/pmodels/argobots/blob/master/src/arch/abtd_env.c#L82
3. `p_global->max_xstreams` is used in mutex, so this mutex can deal with at maximum the total number of cores.
https://github.com/pmodels/argobots/blob/master/src/include/abti_mutex.h#L95
4. However, test uses 4 xstreams.
https://github.com/pmodels/argobots/blob/master/test/basic/mutex_recursive.c#L11

My laptop only has 2 cores, so this benchmark sometime fails.

carns:
Anecdotally I've hit this assertion before too (I don't have a test case handy, I was working on something else at the time and didn't have time to investigate.  I can probably find it again if needed though).

Regardless of the test script settings, in general should this condition cause an assertion?  

I have some use cases where it is helpful to oversubscribe cores because I expect some execution streams to be blocked on system calls to I/O devices, and I need more I/O concurrency than I have cores available.

shintaro-iwasaki:
In general, users should not observe this assertion. This assertion should happen because of this problem.

I agree on the usefulness of oversubscribing cores, but the current Argobots uses by default the number of cores (observed by OS) as ``p_global->num_cores``. 

The workaround is setting an environmental variable (``ABT_MAX_NUM_XSTREAMS``) outside the program, or setting it inside the program before ``ABT_init()`` (e.g., use ``putenv``, see #35 ).

I come up with a few options to address this problem in future.
1. Change the mechanism of ``thread_htable`` so that it does not rely on ``p_global->num_cores``. Argobots must be changed.
2. Provide a function to set ``p_global->num_cores``. Argobots and user programs must be changed.

[...]

PR #35, opened 2018-01-30
Tests on a CPU that have few cores

shintaro-iwasaki:
This PR addresses #34 

The test sometimes failed on machines that have few (<4) cores,
because some tests creates more execution streams than CPU cores.

To solve this issue, all benchmarks specify the max number of
execution streams at the beginning.  The environmental variable
(=ABT_MAX_NUM_XSTREAMS) is set before ABT_init() so that Argobots
can use that value.
Note ABT_MAX_NUM_XSTREAMS is originally employed by Argobots.

### DIFF
--- a/test/basic/barrier.c
+++ b/test/basic/barrier.c
@@ -48,7 +48,7 @@ int main(int argc, char *argv[])
     int ret;
 
     /* Initialize */
-    ATS_init(argc, argv);
+    ATS_read_args(argc, argv);
     if (argc < 2) {
         num_xstreams = 4;
         N = 10;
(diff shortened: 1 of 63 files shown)


halimamer:
Wouldn't it better to query Argobots for the max number of ESs and adjust the test? 
instead of
`num_xstreams = DEFAULT_XSTREAMS`
we do
`num_xstreams = max(DEFAULT_XSTREAMS, ABT_xstream_get_num())`

shintaro-iwasaki:
If tests set the maximum number of execution streams, laptops can fully test Argobots. 

Furthermore, quite a few tests need a large modification to change the number of execution streams.
For example, `pool_access` requires 5 execution streams and it is difficult to change it adaptively. 
https://github.com/pmodels/argobots/blob/master/test/basic/pool_access.c

halimamer:
test:jenkins

shintaro-iwasaki:
test:jenkins

PR #36, opened 2018-01-30
Fix a bug of ABT_cond_broadcast.

shintaro-iwasaki:
This PR fixes a bug of ``ABTI_cond_broadcast`` (no corresponding issue)

The previous ``ABT_cond_timedwait`` and ``ABTI_cond_wait`` try to obtain
``p_cond``'s mutex by spinlock, which causes a deadlock.  It often
happens when waiting threads are signaled simultaneously by
``ABT_cond_broadcast``, and a thread taking a lock yield to another
awaken threads.

By replacing spinlock with a normal mutex lock, another awaken
threads can again yield and the thread taking a lock can be resume.

This commit also changed a test to check this behavior.  The old
Argobots will freeze with ``cond_test``, while this commit passes that
test.

### DIFF
--- a/test/basic/cond_test.c
+++ b/test/basic/cond_test.c
@@ -59,6 +59,7 @@ void inc_counter(void *arg)
     ABT_mutex_lock(mutex);
     g_num_incthreads++;
     ABT_cond_wait(broadcast, mutex);
+    ABT_thread_yield();
     ABT_mutex_unlock(mutex);
 }
 
--- a/src/cond.c
+++ b/src/cond.c
@@ -255,7 +255,7 @@ int ABT_cond_timedwait(ABT_cond cond, ABT_mutex mutex,
     ABTU_free(p_unit);
 
     /* Lock the mutex again */
-    ABTI_mutex_spinlock(p_mutex);
+    ABTI_mutex_lock(p_mutex);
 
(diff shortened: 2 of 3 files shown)


test/basic/cond_test.c
```diff
@@ -59,6 +59,7 @@ void inc_counter(void *arg)
     ABT_mutex_lock(mutex);
     g_num_incthreads++;
     ABT_cond_wait(broadcast, mutex);
+    ABT_thread_yield();
```
halimamer: what is this `yield` for?

shintaro-iwasaki: To enhance the test.
Without this yield, the original version passes this test.
With this yield, the original does not because of the issue this PR solves.

halimamer:
This PR actually highlights a bigger issue that I noticed some time ago. See the issue I created #43

halimamer:
test:jenkins

shintaro-iwasaki:
test:jenkins

test/basic/cond_test.c
halimamer: Please create a separate commit for this patch

shintaro-iwasaki:
test:jenkins

PR #38, opened 2018-01-30
Fix spin lock

shintaro-iwasaki:
It fixes a potential bug of spin lock (No corresponding issue)

This bug (see the first commit log) is fixed in the first commit, while the second commit is for performance improvement.
The second commit uses __atomic built-ins that usually perform better than the naive (manual) spin lock implementation because of compiler supports. In addition, we do not need to pay attention to the location of memory barriers in ``ABTI_spinlock_releas.e``

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -603,6 +603,22 @@ else
     enable_simple_mutex=yes
 fi
 
+
+# check if __atomic_test_and_set() and __atomic_clear() are supported
+AC_TRY_COMPILE(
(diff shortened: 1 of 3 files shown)


halimamer:
@shintaro-iwasaki , let's remove the first commit since it's not 100% correct and it is redundant since the second commit fixes it properly anyway

shintaro-iwasaki:
#42 removes the first commit.
I will close this PR.

Issue #10, opened 2016-10-04
Valgrind support


smseo:
Valgrind is not working with the current Argobots implementation even with --enable-valgrind.

shintaro-iwasaki:
#39 was merged.

PR #39, opened 2018-01-30
Valgrind Support

shintaro-iwasaki:
This problem has been mentioned in #10.

This PR is to support Valgrind.

When`` --enable-valgrind`` is given and compiled with debuging flags,
Argobots registers all stacks allocated from heaps to Valgrind so
that Valgrind can correctly handle stacks used by threads. It is
completely ignored and thus does not affect performance when
Valgrind is disabled (by default), or compiled with debugging flags.


shintaro-iwasaki:
test:jenkins

shintaro-iwasaki:
test:jenkins

shintaro-iwasaki:
test:jenkins

halimamer:
test:jenkins

Issue #14, opened 2017-02-20
Argobots mutex unlock prematurely?


danghvu:
Here is a test case that I think having a race-condition and fail occasionally when I try to cleanup a mutex. I compiled (with gcc 4.9.1) and ran with master branch latest commit.

```bash
while ./hello_world_thread; do :; done
...
[U1:E1] Hello, world!
[U0:E0] Hello, world!
lt-hello_world_thread: thread_htable.c:48: ABTI_thread_htable_free: Assertion `!ret' failed.
Aborted (core dumped)
```

It works with pthread mutex and condition variable (`USE_PTH_MUTEX`). I believe this is due to Argobot still held some internal lock during unlock.

```c
#include <stdio.h>
#include "abt.h"

#define NUM_XSTREAMS    2
// Comment this out to use ABT
// #define USE_PTH_MUTEX

#ifndef USE_PTH_MUTEX
ABT_cond c;
ABT_mutex m;
#else
#include <pthread.h>
pthread_cond_t c;
pthread_mutex_t m;
#endif

volatile int run = 0;
volatile int flag = 0;

void thread_hello(void *arg)
{
    int rank, tid = (int)(size_t)arg;
    ABT_xstream_self_rank(&rank);
    if (tid == 0) {
      while (!run) {};
#ifndef USE_PTH_MUTEX
      ABT_mutex_lock(m);
      flag = 1;
      ABT_cond_signal(c);
      ABT_mutex_unlock(m);
#else
      pthread_mutex_lock(&m);
      flag = 1;
      pthread_cond_signal(&c);
      pthread_mutex_unlock(&m);
#endif
    } else {
#ifndef USE_PTH_MUTEX
      ABT_mutex_lock(m);
      run = 1;
      while (!flag)
        ABT_cond_wait(c, m);
      ABT_mutex_unlock(m);
      ABT_mutex_free(&m);
#else
      pthread_mutex_lock(&m);
      run = 1;
      while (!flag)
        pthread_cond_wait(&c, &m);
      pthread_mutex_unlock(&m);
      pthread_mutex_destroy(&m);
#endif
    }
    printf("[U%d:E%d] Hello, world!\n", tid, rank);
}

int main(int argc, char *argv[])
{
    ABT_xstream xstreams[NUM_XSTREAMS];
    ABT_pool    pools[NUM_XSTREAMS];
    int i;

    ABT_init(argc, argv);

    /* Create ESs */
    ABT_xstream_self(&xstreams[0]);
[...]

PR #41, opened 2018-02-01
[WIP] #14 Fix an assert problem of mutex

shintaro-iwasaki:
This PR addresses #14.

Previously, `p_mutex->p_htable` is not protected by `p_mutex->val` (in other words, `p_mutex` itself.)
#14 reports the following problem:
1. in `ABTI_mutex_unlock`, `p_mutex->p_htable` is used after releasing `p_mutex`. 
2. Another thread takes `p_mutex`, unlock it, and release p_mutex and `p_mutex->p_htable`, though `p_mutex->p_htable` is still used in 1.

To solve this problem, `p_mutex->p_htable` should be protected by `p_mutex`. This PR does so, and solves issues incurred by this change.

### DIFF
--- a/src/include/abti_mutex.h
+++ b/src/include/abti_mutex.h
@@ -51,6 +51,20 @@
         ptr[0] = 0;                                     \
     } while (0)
 
+/* Do not change the values of the constants. */
+
+/* Mutex is unlocked. Anyone not taking a lock tries to change it with CAS. */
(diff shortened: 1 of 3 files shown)


shintaro-iwasaki:
test:jenkins

shintaro-iwasaki:
test:jenkins

PR #42, opened 2018-02-01
Fix spin lock (#38)

shintaro-iwasaki:
This PR reflects the comment in #38. It removes the first commit.

It fixes a potential bug of spin lock (No corresponding issue)
This bug (see the first commit log) is fixed by using __atomic built-ins, which have clear semantics.
They usually perform better than the naive (manual) spin lock implementation because of compiler supports.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -603,6 +603,22 @@ else
     enable_simple_mutex=yes
 fi
 
+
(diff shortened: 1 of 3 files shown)


shintaro-iwasaki:
Remove this PR and submit the improved version as #44.
In #44,  `ABTD_atomic_is_locked_uint32()` is atomic.

PR #44, opened 2018-02-02
Fix spin lock (#42)

shintaro-iwasaki:
This PR is an improvement version of #38 and #42.

The previous spinlock issues mem_barrier "after" releasing locks.

It means that, any memory accesses before membarrier can be reordered, so some memory writes might happen (although it is compiler dependent) after releasing a lock.  Other threads can take that lock and read this memory.

To address this issue, this commit uses GCC built-ins.

As shown in https://gcc.gnu.org/onlinedocs/gcc/_005f_005fatomic-Builtins.html, __atomic_test_and_set() and __atomic_clear() are for spin lock implementation, which usually shows better performance because of compiler help.  In addition, they have clear semantics.

For old GCC, __sync built-ins are supported as well.
https://gcc.gnu.org/onlinedocs/gcc-4.1.2/gcc/Atomic-Builtins.html
However, as written in the first page, new code should always use __atomic rather than __sync.  In this implementation, Argobots uses __sync builtins only when __atomic is not supported.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -603,6 +603,23 @@ else
     enable_simple_mutex=yes
 fi
 
+
+# check if __atomic_test_and_set() and __atomic_clear() are supported
+AC_TRY_COMPILE(
+[#include <stdint.h>],
+[int *lock = 0, val;
+ __atomic_test_and_set((char *)lock, __ATOMIC_ACQ_REL);
+ __atomic_clear((volatile char *)lock, __ATOMIC_RELEASE);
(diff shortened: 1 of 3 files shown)


halimamer:
test:jenkins

halimamer:
test:jenkins

rzambre:
test:jenkins

shintaro-iwasaki:
Will be fixed in #53.

Issue #26, opened 2017-08-24
Fix how units are created and accesses


halimamer:
Currently, units (`ABT_unit`, `ABTI_unit`, or `unit_t`) are manipulated through the right APIs (e.g., `u_create_from_thread`) or accessing directly they internal fields. This is wrong, since those units are exposed to the user, who can create their own structures. They are also internally overloaded, such as for use by synchronization functions outside the intended scope of usage: pool operations.

These issues can be addressed by:
1- Always manipulate units through APIs where necessary 
2- Use different encapsulation methods to build waiting queues in synchronization constructs instead of overloading `ABTI_unit`

PR #45, opened 2018-02-08
WIP: add signal_fn for custom scheduler definition

carns:
Do not merge; this is a prototype for discussion/evaluation.

This pull request adds a new user-definable function pointer to user defined schedulers called signal().  If a scheduler associated with a pool defines this function, then it will be invoked when new work items are added to the pool.

The purpose of this feature is to allow idle schedulers to be awoken when new work arrives.  It's use is demonstrated in a branch of the abt-snoozer scheduler here: https://xgitlab.cels.anl.gov/sds/abt-snoozer/tree/dev-sched-signal.  The original (see origin/master) version of abt-snoozer defines its own pool that stacks on top of the default pool so that it can intercept the push() function for this purpose, which is a little awkward and exposes us to #26.  

The patch does not yet correctly protect data structures in the modified ABTI_pool_retain() and ABTI_pool_release() functions.

The new variation of abt-snoozer is also incomplete; it could be simplified considerably now that it doesn't have to be organized to link a user defined scheduler and user defined pool via their data pointers.

### DIFF
--- a/src/include/abt.h.in
+++ b/src/include/abt.h.in
@@ -290,6 +290,7 @@ typedef void     (*ABT_sched_run_fn)(ABT_sched);
 typedef int      (*ABT_sched_free_fn)(ABT_sched);
 /* To get a pool ready for receiving a migration */
 typedef ABT_pool (*ABT_sched_get_migr_pool_fn)(ABT_sched);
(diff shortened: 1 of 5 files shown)


carns:
Closing this PR.  Will open a new one to consider based on feedback from @halimamer .

PR #46, opened 2018-03-06
add blocking_fifo pool type

carns:
This PR adds a new kind of pool called ABT_POOL_BLOCKING_FIFO.  It reuses existing ABT_POOL_FIFO functionality but layers atop it the ability to have the pop() function block briefly if the pool is empty.  This prevents schedulers from busy spinning when no work units are available, which is helpful for use cases like persistent services that would like to idle gracefully.

This PR also includes:
* a minor general optimization to the sched_run function of all existing schedulers to avoid unnecessary pool size() calls in the scheduler loop for the single pool case
* a new test program for the blocking_fifo pool in test/basic/pool_blocking_fifo.c
* new field in the pool struct that can be used internally to inspect a pool's kind
* modifications to the scheduler loop setup in the basic scheduler to optimize it's behavior when the blocking fifo pool is detected (changing event polling behavior, sched_sleep behavior, and avoiding configurations that could lead to deadlock)

Corresponding code that uses (and tests) this PR is in a development branch of Margo: https://xgitlab.cels.anl.gov/sds/margo/tree/dev-blocking-pool/src

For Margo, this change avoids the problem described in #26, eliminates the need for #25, eliminates our dependency on the abt-snoozer https://xgitlab.cels.anl.gov/sds/abt-snoozer, and eliminates our libev dependency.

### DIFF
--- a/src/pool/fifo.c
+++ b/src/pool/fifo.c
@@ -307,7 +307,7 @@ static ABT_unit pool_pop_shared_blocking(ABT_pool pool)
 
     pthread_mutex_lock(&p_data->blocking_mutex);
     unit = pool_pop_shared(pool);
-    if(!unit) {
+    if(unit == ABT_UNIT_NULL) {
         /* timedwait with 100 ms timeout to give the scheduler a chance to
          * wake up and check for events even if no work units arrive.  We
          * intentionally do not loop here.
--- a/src/pool/fifo.c
+++ b/src/pool/fifo.c
@@ -3,16 +3,21 @@
  * See COPYRIGHT in top-level directory.
  */
 
+#include <errno.h>
 #include "abti.h"
 
 /* FIFO pool implementation */
 
 static int      pool_init(ABT_pool pool, ABT_pool_config config);
+static int      pool_init_blocking(ABT_pool pool, ABT_pool_config config);
 static int      pool_free(ABT_pool pool);
+static int      pool_free_blocking(ABT_pool pool);
 static size_t   pool_get_size(ABT_pool pool);
 static void     pool_push_shared(ABT_pool pool, ABT_unit unit);
+static void     pool_push_shared_blocking(ABT_pool pool, ABT_unit unit);
 static void     pool_push_private(ABT_pool pool, ABT_unit unit);
 static ABT_unit pool_pop_shared(ABT_pool pool);
+static ABT_unit pool_pop_shared_blocking(ABT_pool pool);
 static ABT_unit pool_pop_private(ABT_pool pool);
 static int      pool_remove_shared(ABT_pool pool, ABT_unit unit);
 static int      pool_remove_private(ABT_pool pool, ABT_unit unit);
@@ -31,6 +36,8 @@ struct data {
     size_t num_units;
     unit_t *p_head;
     unit_t *p_tail;
+    pthread_mutex_t blocking_mutex;
+    pthread_cond_t blocking_cond;
 };
 typedef struct data data_t;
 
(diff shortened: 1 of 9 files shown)


src/sched/basic.c
```diff
@@ -112,8 +112,7 @@ static void sched_run(ABT_sched sched)
         for (i = 0; i < num_pools; i++) {
             ABT_pool pool = pools[i];
             ABTI_pool *p_pool = ABTI_pool_get_ptr(pool);
-            size_t size = ABTI_pool_get_size(p_pool);
-            if (size > 0) {
+            if (num_pools == 1 || ABTI_pool_get_size(p_pool) > 0 ) {
```
halimamer: How about we completely get rid of this branch? `get_size > 0` and the beginning of `pop` do the same thing; they check that `num_units > 0`. So might as well save on the redundant branch

src/sched/prio.c
```diff
@@ -95,8 +95,7 @@ static void sched_run(ABT_sched sched)
         for (i = 0; i < num_pools; i++) {
             ABT_pool pool = p_pools[i];
             ABTI_pool *p_pool = ABTI_pool_get_ptr(pool);
-            size_t size = ABTI_pool_get_size(p_pool);
-            if (size > 0) {
+            if (num_pools == 1 || ABTI_pool_get_size(p_pool) > 0) {
```
halimamer: same comment as above

src/sched/randws.c
```diff
@@ -82,8 +82,7 @@ static void sched_run(ABT_sched sched)
         /* Execute one work unit from the scheduler's pool */
         ABT_pool pool = p_pools[0];
         ABTI_pool *p_pool = ABTI_pool_get_ptr(pool);
-        size_t size = ABTI_pool_get_size(p_pool);
-        if (size > 0) {
+        if (num_pools == 1 || ABTI_pool_get_size(p_pool) > 0) {
```
halimamer: same comment as above with a slight variation.
We `pop` a unit, and if that unit is `NULL` then work-steal.

src/pool/fifo.c
```diff
@@ -169,6 +223,21 @@ static void pool_push_shared(ABT_pool pool, ABT_unit unit)
...
+static void pool_push_shared_blocking(ABT_pool pool, ABT_unit unit)
+{
+    ABTI_pool *p_pool = ABTI_pool_get_ptr(pool);
+    void *data = ABTI_pool_get_data(p_pool);
+    data_t *p_data = pool_get_data_ptr(data);
+
+    /* hold additional pthread mutex and signal anyone who might be blocking
+     * on pop()
+     */
+    pthread_mutex_lock(&p_data->blocking_mutex);
```
halimamer: I don't think acquiring the lock here is necessary. The pool push operation should be thread safe and `cond_signal` does not need a lock

src/sched/basic.c
carns: The problem is that the new pool added by this PR (called blocking_fifo in this PR, but possibly to be renamed re: offline discussion) can block on the pop() function.  If you have more than one pool associated with a scheduler then the other pool could be starved if you are using that pool.

I could rework it so that the existing schedulers have the optimization that you describe, and then add a new scheduler that is aware of the blocking behavior and responds accordingly.  We already have to start a new scheduler to swap pools anyway, so it could just as well be a special scheduler intended for use with the blocking_fifo pool.

That's a little bit of code duplication, but it would let us cut down the code path for the default schedulers.

src/pool/fifo.c
carns: I agree cond_signal() itself doesn't need a lock, but the lock in this specific example is to prevent a race between the waiter checking for the pool being empty and this function adding something to the pool.  Without the mutex the you could theoretically get this order of operations, with thread a calling pop() and thread b calling push():

thread a: lock 
thread a: check if pool is empty
thread b: insert into pool queue
thread b: cond_signal
thread a: cond_wait

Thread a will block in this case even though there is work in the pool when it hits the cond_wait() call.

halimamer: the timeout != infinite avoids deadlocks in this PR, so I think it's still correct. Performance wise, my suggestion allows push() to be on a fast path and pop() on empty (including false positives) on a slow path. But I agree that if we change the semantic of the pool to include infinite wait on pop() (which I think is a likely API extension in the future), then keeping the lock for correctness is a must

carns:
Closing this PR.  Will generate some new PRs with a refined design based on offline discussions.

PR #47, opened 2018-03-18
remove pool size check from scheduler inner loops

carns:
the pop() function itself will return NULL if no work units available


halimamer:
test:jenkins

PR #48, opened 2018-03-18
remove extraneous pool_def struct in fifo pool

carns:


carns:
this is included in #49 with better commit log, closing this separate PR

PR #49, opened 2018-03-18
implement ABT_SCHED_BASIC_WAIT

carns:
This adds a new predefined scheduler type called ABT_SCHED_BASIC_WAIT that will block when idle if used in conjunction with ABT_POOL_FIFO_WAIT.

Includes test program and minor change to sched_create_basic() to make it use pool of type POOL_FIFO_WAIT by default with SCHED_BASIC_WAIT if the caller does not specify an explicit pool to use.

This is the revised version of PR #46.  This PR includes generic improvements from PR #47 and PR #48 .

Corresponding branch of margo that uses this capability can be found at https://xgitlab.cels.anl.gov/sds/margo/tree/dev-fifo-wait

### DIFF
--- a/src/pool/fifo.c
+++ b/src/pool/fifo.c
@@ -26,25 +26,6 @@ static ABT_unit unit_create_from_thread(ABT_thread thread);
 static ABT_unit unit_create_from_task(ABT_task task);
 static void unit_free(ABT_unit *unit);
 
-
-/* FIXME: do we need this? */
-ABT_pool_def ABTI_pool_fifo = {
-    .access               = ABT_POOL_ACCESS_MPSC,
-    .p_init               = pool_init,
-    .p_free               = pool_free,
-    .p_get_size           = pool_get_size,
-    .p_push               = pool_push_shared,
-    .p_pop                = pool_pop_shared,
-    .p_remove             = pool_remove_shared,
-    .u_get_type           = unit_get_type,
-    .u_get_thread         = unit_get_thread,
-    .u_get_task           = unit_get_task,
-    .u_is_in_pool         = unit_is_in_pool,
-    .u_create_from_thread = unit_create_from_thread,
-    .u_create_from_task   = unit_create_from_task,
-    .u_free               = unit_free,
-};
-
 struct data {
     ABTI_spinlock mutex;
     size_t num_units;
@@ -57,7 +38,6 @@ static inline data_t *pool_get_data_ptr(void *p_data)
 {
     return (data_t *)p_data;
 }
-
 
 /* Obtain the FIFO pool definition according to the access type */
 int ABTI_pool_get_fifo_def(ABT_pool_access access, ABT_pool_def *p_def)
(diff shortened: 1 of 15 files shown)


halimamer:
test:jenkins

halimamer:
@carns , a minor issue, please look at the Jenkins failures, the builds cannot find `clock_gettime`. 
A more serious issue is with the type of the timeout value. Argobots already has a portability layer for timers (`ABT_timer`). The user should use this type for the timeout value. Internally we can convert the timeout value.

For future reference, please write the comment `test:jenkins` here and it will trigger the Jenkins jobs. Only when the checks are green then I will start a thorough review.

halimamer:
test:jenkins

carns:
@halimamer I might need some clarification on exactly what type to use.  ABT_timer doesn't look quite right because it is (literally) a timer with functions to start and stop the timer and measure elapsed time.  I think what we need here is a way to specify a single point in time.  

The ABTD_time type looks close (and clearly abstracts clock_gettime() for systems that don't have it or don't use it), but it only has 3 accessor functions: _init(), _get(), and _read_sec().  

Should I use that, and add an additional accessor function in arch/abtd_time.c that's similar to _read_sec() except instead of retrieving the value it adds the specified double seconds to the value?  Maybe also an accessor function that does the conversion to struct timespec for pthread_cond_timedwait() so that the conversion is consistent in case anyone else needs to do that?

At any rate, resolving this will fix the clock_gettime() dependency problem, I think.  What's the best way to see the errors that come out of the automated tests?  The details link above on the build test is a little confusing.

carns:
Thanks, that makes sense.  PR has been updated with a commit addressing that.

carns:
test:jenkins

src/pool/fifo.c
```diff
@@ -58,7 +39,6 @@ static inline data_t *pool_get_data_ptr(void *p_data)
     return (data_t *)p_data;
 }
 
-
```
halimamer: This whitespace cleanup is unnecessary

halimamer:
I think the PR looks good. We just need to fix minor `git` cosmetic issues before merging. We will add some of the rules below to our wiki as part of our `git` workflow.

- Commit messages need to define the scope, such as `pool: <rest of commit message>`, `sched: <rest>`, when the commit concerns pool or scheduler operations.
- Can you `rebase` instead of `merge` (commit f9184aa8086be84bc54d1834a86487de9a70057b should not show up)? At this point, we are still following a linear history with `rebase`. I don’t know how long or if we will keep it up
- Whitespaces show up in commits  2729cb545a1c3a5 (file `src/pool/fifo_wait.c`) and f44a8e28bdeed5940a (in file `src/sched/basic_wait.c`)
- I think it’s better to merge the content of commit c0c415cb1b94cd to commits c0037ab74eeae0e7 and 37271a4de74aac
- Commit 4a84e66ed8544db should be merged to 37271a4de74aac65e

PR #50, opened 2018-03-30
Add missing error string in ABT_error_get_str

bcumming:
Add missing error string for `ABT_ERR_INV_MUTEX_ATTR` to  the lookup
table for error strings.

This fixes a bug that lead to:
 * segmentation fault when `ABT_error_str` is called with `err==ABT_ERR_FEATURE_NA`
 * incorrect error strings for `err>ABT_ERR_INV_MUTEX`

### DIFF
--- a/src/error.c
+++ b/src/error.c
@@ -49,6 +49,7 @@ int ABT_error_get_str(int err, char *str, size_t *len)
         "ABT_ERR_INV_TASK",
         "ABT_ERR_INV_KEY",
         "ABT_ERR_INV_MUTEX",
+        "ABT_ERR_INV_MUTEX_ATTR",
         "ABT_ERR_INV_COND",
         "ABT_ERR_INV_RWLOCK",
         "ABT_ERR_INV_EVENTUAL",


bcumming:
This fix adds the missing string, `ABT_ERR_INV_MUTEX_ATTR` to the lookup table.

However, to guard against segmentation faults in the future if the table is out of date,
the range check:

```
    ABTI_CHECK_TRUE(err >= ABT_SUCCESS && err <= ABT_ERR_FEATURE_NA,
                    ABT_ERR_OTHER);
```

could check against the range of the array:

```
     ABTI_CHECK_TRUE(err >= 0 && err <= sizeof(err_str)/sizeof(char*)-1,                                                                                                
                     ABT_ERR_OTHER);
``` 

This would ensure you don't ever go out of bounds.

halimamer:
test:jenkins

halimamer:
@bcumming , thanks for contributing the patch. I will merge it in a minute.
Regarding the array out of bounds issue, as long as the array and macros are kept in sync, both checks are equivalent. If a mismatch exists, even checking the array bounds might not be enough since the macro values might be inconsistent. I am looking into using an `enum` instead of macros. This will allow checking for the bounds of the array to match the number of distinct elements of the enum as well as out of bounds accesses.

carns:
FYI, in the future it might be worth looking into X macros for this sort of thing.

We use them for a similar purpose in Darshan (e.g. https://xgitlab.cels.anl.gov/darshan/darshan/blob/master/darshan-log-format.h#L123) when we need to keep some arrays in sync.

halimamer:
Thank you for the suggestion @carns , I will look into it.

PR #54, opened 2018-04-23
implement ABT_SCHED_BASIC_WAIT (revised)

carns:
This adds a new predefined scheduler type called ABT_SCHED_BASIC_WAIT that will block when idle if used in conjunction with ABT_POOL_FIFO_WAIT.

Includes test program and minor change to sched_create_basic() to make it use pool of type POOL_FIFO_WAIT by default with SCHED_BASIC_WAIT if the caller does not specify an explicit pool to use.

This is the revised version of PR #49. This PR includes generic improvements from PR #47 and PR #48 .

Corresponding branch of margo that uses this capability can be found at https://xgitlab.cels.anl.gov/sds/margo/tree/dev-fifo-wait


carns:
test:jenkins

PR #63, opened 2018-09-10
Inline fpu check

shintaro-iwasaki:
This PR improves performance.

Embed FPU preservation check into assembly code to reduce instructions.

Previously, `preserve_fpu` flag was passed to `jump_fcontext` and `take_fcontext` functions and checked in the assembly functions. `preserve_fpu` is compile-time constant, but `xxx_fcontext` functions cannot be inlined because they are written in assembly, which incurs a runtime overhead.

Now the macro (`#if ABTD_FCONTEXT_PRESERVE_FPU`) is embedded into the assembly code, so there is no runtime overhead. This optimization reduces at least 6 instructions (including 3 branches) per fork-join operation.

### DIFF
--- a/src/arch/fcontext/jump_arm64_aapcs_elf_gas.S
+++ b/src/arch/fcontext/jump_arm64_aapcs_elf_gas.S
@@ -51,6 +51,8 @@
  *                                                     *
  *******************************************************/
 
+#include "abt_config.h"
+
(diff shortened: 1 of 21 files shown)


shintaro-iwasaki:
test:jenkins

shintaro-iwasaki:
test:jenkins

PR #57, opened 2018-05-07
Minor fix / improvement of pop_wait and pop_timedwait

shintaro-iwasaki:
1. Remove `p_pop_wait`
 `p_pop_wait` is a blocking operation, which easily causes a deadlock in a scheduler. Since the same thing can be done by `pop_timedwait` with long wait time. To avoid a potential problem, `p_pop_wait` is removed.
2. Implement the `FIFO` version of `p_pop_timedwait`.
A new pool feature `p_pop_timedwait` was not supported by a `FIFO`-type pool.
This PR implements this feature with `nanosleep`, which is not as efficient as `FIFO_WAIT`, but can support the equivalent functionality.
3. Expose an interface to use `p_pop_wait`.
Add  `ABT_pool_pop_timedwait`, which is necessary to implement a custom scheduler with the new feature.
4. Minor fixes.
Fix `.gitignore`.
Remove trailing whitespaces. in `pool.c`.

PR #64, opened 2018-09-13
FIFO: Fix a bug and warnings.

shintaro-iwasaki:
Fix a bug and warnings which have been introduced in #57.


shintaro-iwasaki:
test:jenkins

PR #65, opened 2018-09-13
Simplify ABT_thread_xxx_many functions.

shintaro-iwasaki:
Remove restrictions of `ABT_thread_create_many`, `ABT_thread_join_many`, and `ABT_thread_free_many` by removing an experimental join_many feature.

These functions create, join and free many threads, but the current implementation only works under very limited circumstances. This was a note of `ABT_thread_join_many`.

```
NOTE: The current implementation is very experimental. It works only for the
case where all ULTs in thread_list are mapped to the same ES as that of the
caller. They do not need to be in the same pool, but pools associated with
target ULTs should be exclusively associated with the caller's ES.
```

They also have an interoperability issue; threads created by `ABT_thread_create` cannot be joined by `ABT_thread_join_many` and vice versa.

This commit still keeps the functions but removes all the restrictions by abolishing the join many optimization. The current implementation is not mature, but there is plenty of room for performance improvement; not only reducing function calling overheads, but also function inlining, efficient joining (in a more general way), and aggregated pool operation (e.g., `push_many`).

This commit does not have any compatibility issue.

### DIFF
--- a/src/arch/abtd_thread.c
+++ b/src/arch/abtd_thread.c
@@ -66,46 +66,6 @@ static inline void ABTD_thread_terminate(ABTI_thread *p_thread)
 #if defined(ABT_CONFIG_USE_FCONTEXT)
     ABTD_thread_context *p_fctx = &p_thread->ctx;
 
-    /* Now, the ULT has finished its job. Terminate the ULT. */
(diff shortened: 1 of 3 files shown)


shintaro-iwasaki:
test:jenkins

shintaro-iwasaki:
test:jenkins

PR #71, opened 2018-09-19
Add ABTU_memcpy and ABTU_memset

shintaro-iwasaki:
This PR adds `ABTU_memcpy` and `ABTU_memset`, which are replaced with GCC `__builtin` functions
if available. ` __builtin` functions can save function calling overheads (and be inlined), so they
perform better especially when the target memory size is small.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -714,6 +714,22 @@ AC_CHECK_LIB(pthread, pthread_join)
 # check pthread_barrier
 AC_CHECK_FUNCS(pthread_barrier_init)
 
+# check builtin functions
+AC_TRY_COMPILE(
+[#include <string.h>], [__builtin_memcpy(NULL, NULL, 1);],
+[have_builtin_memcpy=yes], [have_builtin_memcpy=no]
+)
+AS_IF([test "x$have_builtin_memcpy" = "xyes"],
+      [AC_DEFINE(ABT_CONFIG_HAVE___BUILTIN_MEMCPY, 1,
+                 [Define to 1 if you have the `__builtin_memcpy' function.])])
+AC_TRY_COMPILE(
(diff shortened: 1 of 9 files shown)


shintaro-iwasaki:
test:jenkins

shintaro-iwasaki:
test:jenkins

raffenet:
@shintaro-iwasaki does the compiler not already do this? Or are there configuration options that prevent it?

raffenet:
For instance, I am used to seeing programs built with Intel compilers substitute `_intel_fast_memcpy`. I wonder if GCC will do something similar.

raffenet:
https://gcc.gnu.org/onlinedocs/gcc/Other-Builtins.html#Other-Builtins

According to that page, these functions are by default builtins. Playing around with a toy program, the function call disappears once you enable at least `-O1` optimization level.

shintaro-iwasaki:
Thank you for your comments! You are right. There is no difference between `__builtin_memxxx` abd `memxxx` with widely-used optimization flags.


I checked the behavior of `memxxx` and `__builtin_memxxx` with icc 17.04, clang 6.0, and gcc 4.8.
At least, with `-O3`, they emit exactly the same assembly code.

```
void ext_memcpy(void *a, const void *b, size_t s) {
        memcpy(a, b, s);
}

void ext_memcpy_size(void *a, const void *b) {
        memcpy(a, b, 16);
}

void ext_memset(void *a, char c, size_t s) {
        memset(a, c, s);
}

void ext_memset_size(void *a, char c) {
        memset(a, c, 16);
}
```
All the three compilers above do...
1. If the size is compile-time constant, it inlines `memxxx`
2. If the size is not compile-time constant, it calls `memxxx` functions (i.e., the standard `memset`). Intel calls `_intel_fast_memcpy` and `_intel_fast_memset`.

PR #73, opened 2018-09-20
Add likely and unlikely

shintaro-iwasaki:
This PR adds `ABTU_likely` and `ABTU_unlikely` functions, which give hints to the compiler with branch prediction information. Those features are currently not used. They will be added in future.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -736,6 +736,15 @@ if test -z "$timer_type" ; then
 fi
 AC_MSG_NOTICE([Timer type selected is $timer_type])
 
+# check __builtin_expect
+AC_TRY_COMPILE(
+[], [int cond = 0; if (__builtin_expect(cond, 1)) cond = 2;],
+[have_builtin_expect=yes], [have_builtin_expect=no]
+)
+AS_IF([test "x$have_builtin_expect" = "xyes"],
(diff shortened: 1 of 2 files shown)


configure.ac
```diff
@@ -736,6 +736,15 @@ if test -z "$timer_type" ; then
 fi
 AC_MSG_NOTICE([Timer type selected is $timer_type])
 
+# check __builtin_expect
+AC_TRY_COMPILE(
```
raffenet: FYI, `AC_TRY_COMPILE` is "obsolete" according to the autoconf manual. `AC_COMPILE_IFELSE` should be used instead. https://www.gnu.org/software/autoconf/manual/autoconf-2.69/html_node/Obsolete-Macros.html

shintaro-iwasaki: Thank you! The current `configure.ac` is based on `AC_TRY_COMPILE`. However, it is obsolete even in 2.6.3 (`AC_PREREQ(2.63)` is specified in `configure.ac`). I will check obsolete configure macros and update them in a separate PR if needed.

raffenet: 👍

shintaro-iwasaki:
test:jenkins

PR #72, opened 2018-09-20
Add a new type of ULT

shintaro-iwasaki:
This PR adds a new type of ULT, which performs good when the ULT does not suspend.
Specifically, it implements "return-on-completion" in the following paper:

Lessons Learned from Analyzing Dynamic Promotion for User-Level Threading
S. Iwasaki, A. Amer, K. Taura, and P. Balaji (SC '18)

This PR does not contain "lazy stack allocation", which will be added in the future PR.

PR #74, opened 2018-10-16
[Improvement] Add stacktype

shintaro-iwasaki:
This commit replaces `userstack` of `ABTI_thread_attr` with a new variable `stacktype`.
This PR is a prerequisite of #72.


shintaro-iwasaki:
test:jenkins

src/include/abti_mem.h
```diff
@@ -108,26 +104,22 @@ static inline
...
     size_t stacksize, actual_stacksize;
     char *p_blk;
     ABTI_thread *p_thread;
-    ABTI_stack_header *p_sh;
     void *p_stack;
 
     /* Get the stack size */
     stacksize = *p_stacksize;
-    actual_stacksize = stacksize - header_size;
+    actual_stacksize = stacksize - sizeof(ABTI_thread);
```
halimamer: how is this change, and the several other line changes above and below, is related to `ABTI_EXT_STACK`?
It seems memory allocation is made independent of the header size but it's unclear how it relates to `ABTI_EXT_STACK`

shintaro-iwasaki: `header_size` was `sizeof(ABTI_stack_header) + sizeof(ABTI_thread)`.
This `ABTI_stack_header` is only used to store `ABTI_EXT_STACK`. However, it is not needed.
We don't need alias (i.e., `header_size`) any more and just use `sizeof(ABTI_thread)` directly.

shintaro-iwasaki:
test:jenkins

PR #84, opened 2019-02-08
Fix ABTD_atomic_pause.

shintaro-iwasaki:
Non x86-64 machines fail to compile Argobots, because `ABTD_atomic_pause()` uses an x86/64 specific instruction, `pause`.
This commit removes it when non x86-64 machines are used.

This change passes a quick test on an ARM 64 machine.


halimamer:
test

halimamer:
test:jenkins

PR #88, opened 2019-02-15
Fix a potential bug in thread joining.

shintaro-iwasaki:
A while loop busy in `ABT_thread_join` waits until `u_is_in_pool` returns `ABT_TRUE`, though there is no guarantee that `u_is_in_pool`, which can be given by a user, is aware of memory ordering.

This patch adds a memory barrier to fix this issue.  Note that this path is rarely taken so the additional barrier does not affect performance.


shintaro-iwasaki:
test:jenkins

shintaro-iwasaki:
test:jenkins

PR #83, opened 2019-02-05
Thread safer  ABT_init/finalize/initialized

shintaro-iwasaki:
`ABT_init`, `ABT_finalize`, and `ABT_initialized` do not work well if multiple thread instance tries to call these functions. This situation often happens when multiple Argobots-based systems are employed.

This PR makes them thread safer.

PR #89, opened 2019-02-15
Remove an unused function.

shintaro-iwasaki:
#83 contains an unused function by mistake. This PR removes it.


shintaro-iwasaki:
This has been solved by #123

PR #91, opened 2019-02-16
Implement a dynamic promotion thread for ARM64.

shintaro-iwasaki:
This PR implements `init_and_call_fcontext` for `arm64-elf`, which is a key assembly function for dynamic promotion technique (#72).
Now `x86_64-elf` and `arm64-elf` support the dynamic promotion technique.

It passed all the tests on an ARM machine.


shintaro-iwasaki:
test:argobots/all

shintaro-iwasaki:
I tested it on 64-bit ARM and it worked. This PR should be okay.

PR #95, opened 2019-02-22
Fix #94

### LINKED ISSUE #94: [Bug] memory error when SCHED_STACKSIZE = THREAD_STACKSIZE
When `ABT_THREAD_STACKSIZE` is equal to `ABT_SCHED_STACKSIZE`, the stack of the scheduler ULT (e.g., the primary main scheduler's ULT) uses a stack managed by the memory pool, even though `ABTI_STACK_TYPE_MALLOC` is specified at https://github.com/pmodels/argobots/blob/master/src/thread.c#L1868.
Any attempts to free this scheduler actually free a stack managed by the memory pool, which causes `SEGV` in `free()`  (by freeing illegal memory region) or in the memory pool code afterwards (by freeing a memory pool, not a thread).

The workaround is using different stack size; `ABT_SCHED_STACKSIZE` is by default 4MB, so `ABT_THREAD_STACKSIZE=$(4 * 1024 * 1204 - 64)` actually works.
This bug should be fixed soon.

It has been initially reported by @carns. Thank you very much!

shintaro-iwasaki:
As reported in #94, this bug happens because `malloc` is not used  in `ABTI_mem_alloc_thread()` even when `ABTI_STACK_TYPE_MALLOC` is passed. This PR fixes this bug.

### DIFF
--- a/src/include/abti_mem.h
+++ b/src/include/abti_mem.h
@@ -179,7 +179,8 @@ ABTI_thread *ABTI_mem_alloc_thread(ABTI_thread_attr *p_attr)
         }
 
         stacksize = p_attr->stacksize;
-        if (stacksize != def_stacksize) {
+        if (stacksize != def_stacksize ||
+            p_attr->stacktype == ABTI_STACK_TYPE_MALLOC) {
             /* Since the stack size requested is not the same as default one,
              * we use ABTU_malloc. */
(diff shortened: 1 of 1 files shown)


shintaro-iwasaki:
test:jenkins

yfguo:

PR #99, opened 2019-05-09
Print warning when ABT_MAX_NUM_XSTREAMS is too small

shintaro-iwasaki:
This is an interim solution mentioned by @carns in #34.

This patch is simple; Argobots prints warning (only once) if # of ESs > `ABT_MAX_NUM_XSTREAMS` so that users can set a larger number to `ABT_MAX_NUM_XSTREAMS` (i.e., `ABT_MAX_NUM_XSTREAMS=XXX ./a.out`, or to do it in a program, see #35).

This is a real pain in the neck; currently the priority-/locality-aware mutex implementation (specifically `ABTI_thread_htable` uses a preallocated array whose size is `ABT_MAX_NUM_XSTREAMS` (by default, the number of cores), so any `ABT_mutex` created before creating all the execution streams can cause SEGV if the total number of executions streams happen to be larger than the number of cores). Many users complained about this annoying behavior.

Note that another possible option is returning an error in `ABT_xstream_create`, but this might be too much regarding compatibility since Argobots works (and worked) in the other cases.

Better mutex implementation which does not rely on `ABT_MAX_NUM_XSTREAMS` is future work; it is not very trivial, and importantly we don't know the current mutex implementation is good.

### DIFF
--- a/src/global.c
+++ b/src/global.c
@@ -269,6 +269,26 @@ void ABTI_global_update_max_xstreams(int new_size)
 
     ABTI_spinlock_acquire(&gp_ABTI_global->xstreams_lock);
 
+    static int max_xstreams_warning_once = 0;
+    if (max_xstreams_warning_once == 0) {
+        /* Because some Argobots functionalities depend on the runtime value
+         * ABT_MAX_NUM_XSTREAMS (or gp_ABTI_global->max_xstreams), changing this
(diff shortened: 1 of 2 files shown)


shintaro-iwasaki:
test:jenkins

yfguo:

shintaro-iwasaki:
test:jenkins

Issue #81, opened 2018-12-15
ARM build fails due to ABT_CONFIG_USE_FCONTEXT being unset


dorier:
Same issue as https://github.com/pmodels/argobots/issues/18 and https://github.com/pmodels/argobots/issues/11 but for ARM platforms. The previous issues have been closed a while ago. Any plan to add full support for ucontext when fcontext isn't available?

shintaro-iwasaki:
Thank you for your report. Since it causes several compilation issues, we will support the ucontext version soon.

Apart from the ucontext support that should be done anyway in the near future, ucontext is slow and thus used as a backup method. fcontext should support major architectures such as x86/64, ARM, and Power. Previously the architecture detection failed (#18). I assume that 64-bit ARM should return `aarch64`, unless your environment uses special/new architecture and/or ABI (i.e., calling convention).

Can you run `lscpu` and tell me "Architecure" (which should be equal to `$host_cpu` in the `configure` scripts)?

mdorier:
There seems to be some support for fcontext in ARM now in Boost:
https://github.com/boostorg/context/tree/develop/src/asm
Maybe time to integrate it in Argobots?

mdorier:
Note that my issue earlier was with ARM 32bits (Rasbian on Raspberry Pi), not aarch64.

shintaro-iwasaki:
Oh, I see. Now we don't have ARM 32 bits supports.
We cannot just simply copy the Boost codes to Argobots, we need to customize them somehow (e.g., making `take_fcontext`.) Unfortunately, ARM 32 bits is not a major architecture in HPC, so the support is of low priority.

If you need it beyond your pure curiosity, we/I might work on it; if this is the case, please let me know.

JFYI, #100 will revive the ucontext support, so with this, I hope you can run Argobots on Raspberry Pi

mdorier:
It's more of a curiosity/convenience thing; I'm sometimes using a couple of Raspberry Pi to test out distributed code, and I thought since Boost now supports ARM32 it might be easier than bringing back to ucontext. But if the ucontext works, then great!

shintaro-iwasaki:
ucontext is supported by #100. I haven't tested, but I hope Argobots now works on 32-bit ARM.

shintaro-iwasaki:
Since ucontext is now supported, Argobots should work on 32-bit ARM.
[...]

Issue #11, opened 2016-12-09
KNL cross-compile fails due to ABT_CONFIG_USE_FCONTEXT being unset


kevin-harms:
I was trying to build master on Theta setting
./configure --prefix=<path> --host x86_64 CFLAGS="-03 -xMIC-AVX512"
the configure complete but the build fails with:

  CC       libabt_la-barrier.lo
In file included from ../src/include/abtd.h(48),
                 from ../src/include/abti.h(126),
                 from barrier.c(6):
../src/include/abtd_thread.h(162): error: #error directive: "Not implemented yet"
  #error "Not implemented yet"

Removing the KNL specific option and --host parameter will allow the build to complete.

shintaro-iwasaki:
it is needed to specify the host OS as well. For example, in your case,
```
./configure --prefix=[...] --host x86_64-pc-linux-gnu --enable-fast=O3 CFLAGS="-xMIC-AVX512"
```
works with the latest Argobots and Intel C Compiler.

This issue is due to failure of obtaining `host_os` in `configure`. It is necessary to determine the assembly version of `fcontext`. `ucontext` should be a fallback, but "not implemented yet" at this moment (see also #18)

PR #100, opened 2019-05-10
Support ucontext

shintaro-iwasaki:
This PR provides (or revives) `ucontext` support, which is a fallback method when `fcontext` is unavailable.  The previous `ucontext` support was incomplete: compilation fails by `#error`. Though `fcontext` supports many platforms, this PR can widen the support architectures.

The first two commits remove the current incomplete and dirty `ucontext` implementation, and the third commit creates thin wrappers for `fcontext`, and the fourth commit implements their `ucontext` version. The last commit is clean-up.

This PR is related to #11, #18, and #81, all of which as a result encountered the unsupported `ucontext` error.

### DIFF
--- a/src/arch/abtd_thread.c
+++ b/src/arch/abtd_thread.c
@@ -12,13 +12,13 @@ static inline void ABTD_thread_terminate_sched(ABTI_local *p_local,
 
 void ABTD_thread_func_wrapper_thread(void *p_arg)
 {
-    ABTD_thread_context *p_fctx = (ABTD_thread_context *)p_arg;
-    void (*thread_func)(void *) = p_fctx->f_thread;
+    ABTD_thread_context *p_ctx = (ABTD_thread_context *)p_arg;
+    void (*thread_func)(void *) = p_ctx->f_thread;
 
-    thread_func(p_fctx->p_arg);
+    thread_func(p_ctx->p_arg);
 
     /* NOTE: ctx is located in the beginning of ABTI_thread */
-    ABTI_thread *p_thread = (ABTI_thread *)p_fctx;
+    ABTI_thread *p_thread = (ABTI_thread *)p_ctx;
 #ifndef ABT_CONFIG_DISABLE_STACKABLE_SCHED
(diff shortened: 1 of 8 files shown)


mdorier:
I tried out this PR on a 32bits ARMv7 (which doesn't have fcontext support so ucontext is used); I'm getting the following assertion error:

```
thread_htable.c:11: ABTI_thread_htable_create: Assertion `sizeof(ABTI_thread_queue) == 192' failed.
```

shintaro-iwasaki:
@mdorier Thank you for reporting a bug! Argobots does not seem to support the 32-bit architecture. I am making a separate PR for this issue.

This bug happens because `ABTI_thread_htable_create` assumes a specific size of `ABTI_thread_queue` (https://github.com/pmodels/argobots/blob/master/src/thread_htable.c#L11) while it depends on the pointer size (https://github.com/pmodels/argobots/blob/master/src/include/abti_thread_htable.h#L26).

mdorier:
Thanks. There are definitely some aspects of this that I fail to understand (i.e. why the assertion, and why the padding in this structure). Is that easily correctable?

shintaro-iwasaki:
First of all, this `ABTI_thread_htable` is not used by default after https://github.com/pmodels/argobots/pull/102. I guess this size constraint is for performance, but need to investigate it.

shintaro-iwasaki:
test:jenkins/ucontext

shintaro-iwasaki:
test:jenkins
test:jenkins/fcontextdp

shintaro-iwasaki:
test:jenkins/fcontextdp
test:jenkins/ucontext

shintaro-iwasaki:
test:jenkins
test:jenkins/fcontextdp
test:jenkins/ucontext

shintaro-iwasaki:
test:jenkins

shintaro-iwasaki:
test:jenkins
test:jenkins/fcontextdp
test:jenkins/ucontext
test:jenkins/strict

PR #102, opened 2019-06-19
configure: enable simple mutex by default

shintaro-iwasaki:
Currently, Argobots uses an advanced mutex implementation using a hand-off mechanism by default.  However, this mutex algorithm has algorithmic issues.

1. it potentially causes SEGV (#14, #41)
2. it forces users to set the maximum number of execution streams before initialization (#34)

A quick hack does not work for these issues. We need to completely improve the current mutex implementation. It is the right way to fix these issues, but it requires longer time than I expect.

For stability, this PR disables the current experimental mutex implementation and uses the stable simple mutex algorithm, which does not have any reported issues.

Note that this change might negatively affect the mutex performance; one can enable the previous mutex by setting `--disable-simple-mutex`.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -195,9 +195,10 @@ AC_ARG_ENABLE([dynamic-promotion],
                    [use the dynamic promotion threading technique that performs
                     better if a ULT does not yield.]))
 
-# --enable-simple-mutex
+# --disable-simple-mutex
 AC_ARG_ENABLE([simple-mutex],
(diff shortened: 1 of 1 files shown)


halimamer:
@pavanbalaji , somebody should merge this. We left a broken mutex implementation as the default long enough

shintaro-iwasaki:
test:jenkins

shintaro-iwasaki:
test:jenkins

yfguo:

PR #105, opened 2019-07-02
abtu: always use aligned malloc.

shintaro-iwasaki:
Currently, some Argobots codes use `malloc`, `calloc` and `realloc` via `ABTU_malloc` etc in order to allocate data. Since these functions are not aware of cache line alignment. this allocation sometimes causes cache line conflict.

For example,
```
void case1() { // Conflict between Argobots objects
  ABT_mutex mutex1, mutex2;
  ABT_mutex_create(&mutex1); // It internally calls ABTU_calloc, which is mapped to calloc
  ABT_mutex_create(&mutex2); // It internally calls ABTU_calloc, which is mapped to calloc
  // mutex1 and mutex2 can share the same cache lines.
}
void case2() { // Conflict between an Argobot object and user data
  ABT_mutex mutex;
  ABT_mutex_create(&mutex); // It internally calls ABTU_calloc, which is mapped to calloc
  int *user_data = (int *)malloc(sizeof(int) * 16);
  // mutex and user_data can share the same cache line.
}
```
See the implementation of `ABT_mutex_create` for details (See https://github.com/pmodels/argobots/blob/master/src/mutex.c#L36). Note that it actually happened in the past.

To avoid any potential cache line conflict, all the allocation in Argobots should be aligned and the size should be a multiple of the cache line size. This PR improves `ABTU_malloc`, `ABTU_calloc`, and `ABTU_realloc` to return aligned address by rounding up the size to a multiple of cache line size.

Minor changes:
- Since aligned `realloc` is not available, callers need to specify the old size as well (i.e., `ABTU_realloc` takes `old_size`.)
- Replace `ABTU_CA_MALLOC` (cache-aligned malloc) by `ABTU_malloc`.

### DIFF
--- a/src/event.c
+++ b/src/event.c
@@ -800,14 +800,18 @@ int ABT_event_add_callback(ABT_event_kind event,
             max_num = gp_einfo->max_stop_xstream_fn;
             if (cur_num == max_num) {
                 /* We need to allocate more space */
+                size_t cur_size;
                 max_num = max_num * 2;
                 gp_einfo->max_stop_xstream_fn = max_num;
+                cur_size = cur_num * 2 * sizeof(ABT_event_cb_fn);
                 new_size = max_num * 2 * sizeof(ABT_event_cb_fn);
                 gp_einfo->stop_xstream_fn = (ABT_event_cb_fn *)
-                    ABTU_realloc(gp_einfo->stop_xstream_fn, new_size);
+                    ABTU_realloc(gp_einfo->stop_xstream_fn, cur_size, new_size);
+                cur_size = cur_num * 2 * sizeof(void *);
                 new_size = max_num * 2 * sizeof(void *);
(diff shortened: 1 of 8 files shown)


shintaro-iwasaki:
test:jenkins

shintaro-iwasaki:
test:jenkins

shintaro-iwasaki:
test:jenkins

shintaro-iwasaki:
test:jenkins

raffenet:
Node label was missing from Jenkins config. Trying again...
test:jenkins

shintaro-iwasaki:
test:jenkins

shintaro-iwasaki:
test:jenkins

shintaro-iwasaki:
test:jenkins/all

shintaro-iwasaki:
test:jenkins/all

shintaro-iwasaki:
test:jenkins/all

PR #107, opened 2019-08-15
Set up Travis CI for style check

shintaro-iwasaki:
As done by many other projects (including MPICH), Argobots should check the coding style automatically. This PR sets up scripts for Travis CI, which checks trailing white spaces. Note that the original `maint/hook/pre-commit` is that of MPICH (https://github.com/pmodels/mpich/blob/master/maint/hooks/pre-commit).

Specifically, `.travis.yml` executes `maint/hook/pre-commit`, which performs style check by running `maint/code-cleanup.sh`.

The current version of `code-cleanup.sh` is outdated; the future PR should update it.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,13 @@
+sudo: false
+language: c
+
+os: linux
+
+dist: bionic
+
+addons:
+  apt:
+    packages: indent
+
+script:
+  - maint/hook/pre-commit
--- a/maint/hook/pre-commit
+++ b/maint/hook/pre-commit
@@ -1,0 +1,76 @@
+#!/bin/bash
+#
+# An example hook script to verify what is about to be committed.
+# Called by "git commit" with no arguments.  The hook should
+# exit with non-zero status after issuing an appropriate message if
+# it wants to stop the commit.
+#
+# To enable this hook, rename this file to "pre-commit".
+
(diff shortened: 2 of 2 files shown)


raffenet:
> The current version of `code-cleanup.sh` is outdated; the future PR should update it.

I think a lot of projects use `clang-format` these days. I don't know anything more about it, but it would be nice to learn https://clang.llvm.org/docs/ClangFormat.html.

shintaro-iwasaki:
That's a really good point. There are many pros and cons among `indent`, `clang-format` and possibly `astyle`. If `clang-format` is popular enough (and thus everyone is willing to install it), I'd like to use it since it is smarter than the others (at least we don't need to specify typedefs).

The biggest problem of `clang-format` is that it is still evolving; therefore, it is hard to define the stable version (e.g.,  `AlignConsecutiveMacros` (https://clang.llvm.org/docs/ClangFormatStyleOptions.html) is from Clang 9.0 (not released yet!), which is very useful for the C project).

I should more seriously think about how to use `clang-format`.

shintaro-iwasaki:
It looks working properly (#108). The diff in Travis looks good (https://travis-ci.org/pmodels/argobots/jobs/572431243).

Can someone review and merge this? I am going to update the `maint/code-cleanup.sh` (and thus all the other source code) later.

PR #118, opened 2019-09-27
Fix test/basic/info_stackdump2.c

shintaro-iwasaki:
This patch fixes the loop range in `test/basic/info_stackdump2.c` which sometimes causes an illegal memory free error.

The bug is as follows: `args[i][j].stack` is only initialized with `i != 1` (see line 143)
```c
    /* Create ULTs. */
    /* i starts from 1 because one execution stream must be busy-scheduling. */
    for (i = 1; i < num_xstreams; i++) {
        for (j = 0; j < num_threads; j++) {
            [...]
            args[i][j].stack = NULL;
```
while previously it iterates over all `args[i][j].stack`, some of which are uninitialized (see line 212).
```c
    for (i = 0; i < num_xstreams; i++) { // this "0" is the problem
        for (j = 0; j < num_threads; j++) {
            if (args[i][j].stack)
                free(args[i][j].stack);
        }
        [...]
```

### DIFF
--- a/test/basic/info_stackdump2.c
+++ b/test/basic/info_stackdump2.c
@@ -209,11 +209,13 @@ int main(int argc, char *argv[])
     /* Finalize */
     ret = ATS_finalize(0);
 
-    for (i = 0; i < num_xstreams; i++) {
+    for (i = 1; i < num_xstreams; i++) {
         for (j = 0; j < num_threads; j++) {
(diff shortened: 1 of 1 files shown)


shintaro-iwasaki:
test:jenkins

shintaro-iwasaki:
test:jenkins

Issue #119, opened 2019-10-10
memory leak in ABT_init


jolivier23:
If I run the following code under valgrind with --leak-check=full I get a memory leak
int main(int argc, char **argv)
{
	ABT_init(0, NULL);
	ABT_finalize();
	return 0;
}

Here is the call stack for the reported leak.   It doesn't appear to be freeing all memory allocated for main xstream.

==8020== 40 bytes in 1 blocks are definitely lost in loss record 1 of 1
==8020==    at 0x483880B: malloc (vg_replace_malloc.c:309)
==8020==    by 0x485908D: ABTI_elem_create_from_xstream (in /home/jvolivie/prebuilt/argobots/lib/libabt.so.0.0.0)
==8020==    by 0x4852D15: ABTI_xstream_create (in /home/jvolivie/prebuilt/argobots/lib/libabt.so.0.0.0)
==8020==    by 0x4852EAA: ABTI_xstream_create_primary (in /home/jvolivie/prebuilt/argobots/lib/libabt.so.0.0.0)
==8020==    by 0x484DF7E: ABT_init (in /home/jvolivie/prebuilt/argobots/lib/libabt.so.0.0.0)
==8020==    by 0x401153: main (in /home/jvolivie/daos/a.out)

shintaro-iwasaki:
Thank you for reporting a bug. It seems that the bug reported by #6 has not been fixed yet.
I will look into it and hopefully make a patch within 24 hours.

shintaro-iwasaki:
It was fixed by #120.

PR #120, opened 2019-10-10
fix the memory leak

shintaro-iwasaki:
This commit eliminates legacy structs, `ABTI_elem`, `ABTI_contn`, and `ABTI_xstream_contn` from Argobots.  This patch, as a result, solves the memory leak issue reported in #6 and #119; ES's `ABTI_elem` was not only created but not freed (see `src/stream.c`).


shintaro-iwasaki:
test:jenkins

roblatham00:
If you are sure that no users rely on ABTI_elem, ABTI_contn, or ABTI_xstream_contn, then great. (given the names, I guess they are internal routines so end-users using them get what they deserve).

shintaro-iwasaki:
@roblatham00 You're right. They are internal objects, so no users can access them.

roblatham00:
:+1:   ship it!

Issue #111, opened 2019-08-29
Argobots breaks if ICC-19 is used.


shintaro-iwasaki:
If ICC-19 is used, `lp_ABTI_local` is not properly loaded because of #55. This problem could potentially happen with any smart compilers, but it is now confirmed with Intel Compiler.

Specifically, this causes the following error (https://jenkins-pmrs.cels.anl.gov/job/bolt-review-centos/8/, Intel+Argobots `omp_task_nest_tied.c`, `omp_taskloop_grainsize.c`, and `omp_taskloop_num_tasks.c`). It has been confirmed by
1. The SEGV disappears if `#pragma intel optimization_level 0` is put before `ABTI_mem_free_thread()` (all the other functions are compiled with `-O2`). (https://github.com/pmodels/argobots/blob/master/src/include/abti_mem.h#L238)
2. `p_local` is not `lp_ABTI_local` here if the default -O2 optimization flag is passed (checked by inserting `ABTI_ASSERT` + surely truly loaded `lp_ABTI_local`); (https://github.com/pmodels/argobots/blob/master/src/include/abti_mem.h#L248)

I created an independent issue because I confirmed this #55 issue in the major environment in reality: the latest ICC (icc-19.0.3.199) + Intel Xeon (Gold 6140) + CentOS (7) + default compilation flags. I note that this is not a problem of ICC.

This bug will be fixed very soon.

shintaro-iwasaki:
This should be fixed by #124.

Issue #55, opened 2018-04-23
Avoid compiler-level TLS caching.


shintaro-iwasaki:
Argobots internally uses compiler/segment register/OS/`__thread`-level TLS (e.g., `ABTI_local_get_thread`).
Sometimes the OS-level thread running the function is changed across the function call (e.g., `ABTI_xstream_schedule_thread`),
while some compilers (including GCC) assume the same OS-level thread executes this function and thus cache the TLS address.
Argobots does not consider these cases, potentially causing a bug with specific compilers + specific architectures.

Please see the following for this issue.
GCC
https://gcc.gnu.org/bugzilla/show_bug.cgi?id=26461
LLVM
https://bugs.llvm.org/show_bug.cgi?id=19177
MSVC
https://msdn.microsoft.com/en-us/library/6e298fy4.aspx

shintaro-iwasaki:
Addressed by #124

PR #124, opened 2019-11-15
Fix TLS Caching Issue (#55, #111)

shintaro-iwasaki:
This commit fixes the TLS caching issue mentioned in #55 and #111 

## Problem & Cause

Consider the following code:
```c
__thread ABTI_local *lp_ABTI_local;
int ABT_any_func()
{
    ABTI_xstream *p_xstream = lp_ABTI_local->p_xstream;
    ABTI_thread_yield(...);
    ABTI_xstream *p_xstream2 = lp_ABTI_local->p_xstream;
}
```
However, even if xstream is changed before and after `ABTI_thread_yield()`, `p_xstream` can be always equal to `p_xstream2`. This is because major compilers do not assume the running Pthreads can be changed across function calls and thus cache a TLS value (for example, an `fs` register value). In this specific example, the second `lp_ABTI_local` is that of the original `lp_ABTI_local` even if `ABTI_thread_yield()` changes the running Pthreads by user-level context switches.

It can potentially happen with major compilers (GCC, ICC, and Clang) on any architectures (including x86/64, ARM64, ...) as explained in #55. It has been ignored, it became a real problem (#111).

## Solution

This PR creates new functions: `ABT_local_get_local()` and  `ABT_local_get_local_uninlined()`. The first function is fast but might use a cached TLS while the `_uninlined` is slow but does not use a cached TLS.

However, since `_uninlined` is very slow, we need to use `ABT_local_get_local()` as much as possible. To mitigate the performance degradation, this PR adds new rules:
- Use `ABT_local_get_local()` (if needed) at the beginning of the `ABT_` functions and functions that are used as function pointers (e.g., `sched_run`) and thus cannot be inlined.
- Use a `p_local` argument in other functions (most `ABTI_` and `ABTD_` functions) that just read `lp_ABTI_local`.
- Use a `pp_local` argument in functions may change `lp_ABTI_local` internally by context switching (e.g., ABTI_yield).
- Use `ABT_local_get_local_uninlined()` in functions that really change the running Pthreads or are not in the critical path (e.g., `ABTI_log` functions). 

As a result, most `ABT_` functions load TLS with fast `ABT_local_get_local()` and pass `p_local` or `&p_local` to `ABTI_` functions. `ABTI_` and `ABTD_` functions that use `lp_ABTI_local` become taking either `p_local` or `pp_local`. Very few functions use `ABT_local_get_local_uninlined()`.

## Implementation
`ABT_local_get_local()` is an inline function that just reads a TLS directly from the pointer. `ABT_local_get_local_uninlined()` executes a TLS read function (`ABT_local_get_local_internal()`: the same as `ABT_local_get_local()`) via a function pointer stored in a global structure. Since this function pointer can be changed anywhere, the compiler cannot optimize out this function call. `ABT_local_get_local_internal()` cannot use TLS since it cannot be inlined (because it is a function call), so it freshly reads the current TLS.

## Performance Impact
This change is necessary for correctness, but it negatively affects the performance since it potentially adds a TLS read overhead on `ABT_local_get_local_internal()`. This patch also forces functions to pass TLS, which can increase the register pressure. The optimization can be done in the future.

### DIFF
--- a/src/arch/abtd_thread.c
+++ b/src/arch/abtd_thread.c
@@ -5,8 +5,10 @@
 
 #include "abti.h"
 
-static inline void ABTD_thread_terminate_thread(ABTI_thread *p_thread);
-static inline void ABTD_thread_terminate_sched(ABTI_thread *p_thread);
+static inline void ABTD_thread_terminate_thread(ABTI_local *p_local,
+                                                ABTI_thread *p_thread);
+static inline void ABTD_thread_terminate_sched(ABTI_local *p_local,
+                                               ABTI_thread *p_thread);
 
 #if defined(ABT_CONFIG_USE_FCONTEXT)
 void ABTD_thread_func_wrapper_thread(void *p_arg)
@@ -22,7 +24,8 @@ void ABTD_thread_func_wrapper_thread(void *p_arg)
     ABTI_ASSERT(p_thread->is_sched == NULL);
 #endif
 
-    ABTD_thread_terminate_thread(p_thread);
+    ABTI_local *p_local = lp_ABTI_local;
+    ABTD_thread_terminate_thread(p_local, p_thread);
 }
 
 void ABTD_thread_func_wrapper_sched(void *p_arg)
@@ -38,7 +41,8 @@ void ABTD_thread_func_wrapper_sched(void *p_arg)
     ABTI_ASSERT(p_thread->is_sched != NULL);
 #endif
 
-    ABTD_thread_terminate_sched(p_thread);
+    ABTI_local *p_local = lp_ABTI_local;
+    ABTD_thread_terminate_sched(p_local, p_thread);
 }
 #else
 void ABTD_thread_func_wrapper(int func_upper, int func_lower,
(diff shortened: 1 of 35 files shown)


shintaro-iwasaki:
test:jenkins
test:jenkins/32bit
test:jenkins/warning
test:jenkins/strict

shintaro-iwasaki:
test:jenkins
test:jenkins/fcontextdp
test:jenkins/32bit
test:jenkins/warning
test:jenkins/strict

shintaro-iwasaki:
test:jenkins
test:jenkins/fcontextdp
test:jenkins/32bit
test:jenkins/warning
test:jenkins/strict

shintaro-iwasaki:
test:jenkins
test:jenkins/fcontextdp
test:jenkins/32bit
test:jenkins/warning
test:jenkins/strict

shintaro-iwasaki:
test:jenkins
test:jenkins/fcontextdp
test:jenkins/32bit
test:jenkins/warning
test:jenkins/strict

shintaro-iwasaki:
test:jenkins
test:jenkins/fcontextdp
test:jenkins/32bit
test:jenkins/warning
test:jenkins/strict

shintaro-iwasaki:
test:jenkins
test:jenkins/fcontextdp
test:jenkins/32bit
test:jenkins/warning
test:jenkins/strict

shintaro-iwasaki:
test:jenkins
test:jenkins/fcontextdp
test:jenkins/32bit
test:jenkins/warning
test:jenkins/strict

shintaro-iwasaki:
test:jenkins

Issue #93, opened 2019-02-20
Configuration-time huge page size detection for mmap


shintaro-iwasaki:
It is related to #19; Argobots now assumes a certain huge page size (2MB?), but on some machines this size  is different (e.g., some POWER machines use 16MB), which causes an error when we *free* the huge page by `munmap`. Typically it happens in `ABTI_mem_finalize` (in `ABT_finalize`).

Argobots should detect the huge page size in `configure`.

Note that the current workaround is either
- disabling huge page use (`ABT_MEM_LP_ALLOC=malloc`), or
- setting the correct huge page size (e.g., `ABT_MEM_STACK_PAGE_SIZE=$((16*1024*1024))`)

carns:
These environment variables did not work around the problem for me.  I still get crashes here:

```
(gdb) where
#0  0x000010000040d4d8 in ABTD_atomic_fetch_add_uint32 (v=1, 
    ptr=<optimized out>) at ../../src/include/abtd_atomic.h:283
#1  ABTI_mem_free_stack_list (p_stack=<optimized out>)
    at ../../src/mem/malloc.c:212
#2  ABTI_mem_finalize_local (p_local=0x1002375fe60)
    at ../../src/mem/malloc.c:93
#3  0x00001000003f9b64 in ABTI_local_finalize () at ../../src/local.c:42
#4  0x00001000003f85a4 in ABT_finalize () at ../../src/global.c:208
#5  0x00001000002b4654 in hij ()
   from /autofs/nccs-svm1_home1/carns/working/src/spack/opt/spack/linux-rhel7-ppc64le/gcc-6.4.0/margo-0.4.2-oi35bj5ssdltzdy7vyxgaw7j4b7zyfhm/lib/libmargo.so.0
#6  0x0000000010002158 in main (argc=5, argv=0x3fffed202da8)
    at ../perf-regression/margo-p2p-latency.c:161
```

If I comment out https://github.com/pmodels/argobots/blob/master/src/mem/malloc.c#L93 (just completely skip ABTI_mem_free_stack_list() as a quick hack) then my test cases complete 100% successfully with no environment variables set.  This code path seems to be the only issue for my test case at least.

raffenet:
@shintaro-iwasaki I'm not sure configuration-time detection can work. At least, it won't be cross-compile safe, since huge page settings are dependent on OS configuration.

shintaro-iwasaki:
Thank you very much!
[...]

PR #129, opened 2019-11-17
Disable hugepage allocation by default on non-x86/64.

shintaro-iwasaki:
This is a workaround for #93.

When hugepage is used, `mmap()` needs a correct size of hugepage; otherwise, error happens on `munmap()`.  However, the default size is for typical x86/64 machines, not for other architectures, so the error happens when the hugepage size is different.  To run Argobots with default settings on these architectures, we disable hugepage allocation by default on non-x86/64 architectures; on such a machine, hugepage settings should be explicitly enabled via an environmental variable.

In the future, this issue should be fixed by using a correct hugepage size.


shintaro-iwasaki:
test:jenkins

PR #131, opened 2019-12-09
sched: clean up the algorithm of the basic scheduler

pnookala:
This PR changes the algorithm of the basic scheduler to count the number of pops for every pool. 
 It checks for events after X ULTs rather than X passes through all the pools. 

Also,  this PR removed the run_cnt variable used to sleep the scheduler replacing it with a single line of code.


shintaro-iwasaki:
Thank you for the PR! I haven't looked at the code, but would you please rebase your PR and remove unrelated commits?
Also, please follow the format of the commit message.
```
sched: clean up the algorithm of the basic scheduler

(Reason why you change it)
```

> checks for events after X ULTs rather than X passes through all the pools.
This is not an advantage; it just explains the behavior of the new scheduler. Why did you change it?

PR #132, opened 2019-12-09
sched: clean up the algorithm of the basic scheduler

pnookala:
This change to the basic scheduler algorithm checks for events after X ULTs rather than X passes
through all the pools. The core idea is that pop_count is a better metric to check for an event periodically compared with previous work_count (pop_count * num_pools) since this metric does not depend on the number of pools. Also, this change removed run_cnt variable and used unit variable as input for SCHED_SLEEP, which makes the code a bit cleaner since another variable is not required.

### DIFF
--- a/src/sched/basic.c
+++ b/src/sched/basic.c
@@ -91,13 +91,13 @@ static int sched_init(ABT_sched sched, ABT_sched_config config)
 static void sched_run(ABT_sched sched)
 {
     ABTI_local *p_local = ABTI_local_get_local();
-    uint32_t work_count = 0;
+    ABT_unit unit = ABT_UNIT_NULL;
+    uint32_t pop_count = 0;
     sched_data *p_data;
     uint32_t event_freq;
     int num_pools;
     ABT_pool *pools;
     int i;
-    CNT_DECL(run_cnt);
 
     ABTI_xstream *p_xstream = p_local->p_xstream;
     ABTI_sched *p_sched = ABTI_sched_get_ptr(sched);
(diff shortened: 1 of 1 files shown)


src/sched/basic.c
```diff
@@ -107,31 +106,25 @@ static void sched_run(ABT_sched sched)
...
-        CNT_INIT(run_cnt, 0);
-
-        /* Execute one work unit from the scheduler's pool */
-        for (i = 0; i < num_pools; i++) {
-            ABT_pool pool = pools[i];
-            ABTI_pool *p_pool = ABTI_pool_get_ptr(pool);
-            /* Pop one work unit */
-            ABT_unit unit = ABTI_pool_pop(p_pool);
-            if (unit != ABT_UNIT_NULL) {
+        for (int j = 0; j < num_pools; j++) {
```
shintaro-iwasaki: Please use `i` instead of `j` because it is the outermost for loop.
```
for (i = 0; i < num_pools; i++) {
```

src/sched/basic.c
```diff
@@ -107,31 +106,25 @@ static void sched_run(ABT_sched sched)
...
         }
-
-        if (++work_count >= event_freq) {
+        /* if we attempted event_freq pops, check for events */
+        if (pop_count >= event_freq) {
             ABTI_xstream_check_events(p_xstream, sched);
-            ABT_bool stop = ABTI_sched_has_to_stop(&p_local, p_sched,
-                                                   p_xstream);
-            if (stop == ABT_TRUE)
+            if (ABTI_sched_has_to_stop(&p_local, p_sched, p_xstream)
```
shintaro-iwasaki: Remove the trailing whitespace.

src/sched/basic.c
```diff
@@ -107,31 +106,25 @@ static void sched_run(ABT_sched sched)
     event_freq = p_data->event_freq;
     num_pools  = p_data->num_pools;
     pools      = p_data->pools;
+    ABT_unit unit = ABT_UNIT_NULL;
```
shintaro-iwasaki: Please declare it either at the beginning of the function or at the beginning of the `while` loop.

src/sched/basic.c
```diff
@@ -107,31 +106,25 @@ static void sched_run(ABT_sched sched)
...
-
-        if (++work_count >= event_freq) {
+        /* if we attempted event_freq pops, check for events */
+        if (pop_count >= event_freq) {
             ABTI_xstream_check_events(p_xstream, sched);
-            ABT_bool stop = ABTI_sched_has_to_stop(&p_local, p_sched,
-                                                   p_xstream);
-            if (stop == ABT_TRUE)
+            if (ABTI_sched_has_to_stop(&p_local, p_sched, p_xstream) 
+                                                == ABT_TRUE)
```
shintaro-iwasaki: This indent is weird. Please align it to somewhere
```
            if (ABTI_sched_has_to_stop(&p_local, p_sched, p_xstream) 
                == ABT_TRUE)
```

shintaro-iwasaki:
Note the original PR is #131.

src/sched/basic.c
shintaro-iwasaki: I don't think it has been removed. Could you please confirm it again?

pnookala: Could you please clarify which line has the trailing white space?

shintaro-iwasaki:
```
if (ABTI_sched_has_to_stop(&p_local, p_sched, p_xstream)[ ]<=
```

shintaro-iwasaki:
@pnookala
Can you only modify your local commits and push it only when it (log, commits, ...) looks good? It makes the GitHub history a little bit cleaner.

src/sched/basic.c
pnookala: Removed.

shintaro-iwasaki:
As I mentioned in #131, could you please follow the format of the commit message?
```
sched: ***

(Reason why you change it)
```
You can see examples by typing `git log`.
Thanks!

src/sched/basic.c
shintaro-iwasaki: Please fix it as we discussed.

src/sched/basic.c
shintaro-iwasaki: Please fix it as we discussed.

src/sched/basic.c
```diff
@@ -109,29 +109,22 @@ static void sched_run(ABT_sched sched)
...
-        CNT_INIT(run_cnt, 0);
-
-        /* Execute one work unit from the scheduler's pool */
         for (i = 0; i < num_pools; i++) {
-            ABT_pool pool = pools[i];
-            ABTI_pool *p_pool = ABTI_pool_get_ptr(pool);
-            /* Pop one work unit */
-            ABT_unit unit = ABTI_pool_pop(p_pool);
-            if (unit != ABT_UNIT_NULL) {
+            ABTI_pool *p_pool = ABTI_pool_get_ptr(pools[j]);
```
shintaro-iwasaki: Please change it to i.

PR #134, opened 2019-12-10
Clean up ABTI_error.h

shintaro-iwasaki:
This PR cleans up `ABTI_error.h`.

### 1. Clean up the implementation of `ABTI_CHECK_xxx` and `HANDLE_xxx` macros.

Previously, two implementations of `ABTI_CHECK_` and `HANDLE_xxx` macros exist based on `ABT_CONFIG_DISABLE_ERROR_CHECK` and `ABT_CONFIG_PRINT_ABT_ERRNO`. This PR creates constant values associated with `ABT_CONFIG_DISABLE_ERROR_CHECK` and `ABT_CONFIG_PRINT_ABT_ERRNO` and add a branch in the macros (without creating a macro).

Briefly speaking,
```c
#ifndef ABT_CONFIG_DISABLE_ERROR_CHECK
#define ABTI_CHECK_TRUE_RET(cond,val) \
    do {                              \
        if (!(cond)) {                \
            return (val);             \
        }                             \
    } while(0)
#else
#define ABTI_CHECK_TRUE_RET(cond,val)
#endif
```
becomes the following.
```c
// ABT_CONFIG_IS_ERROR_CHECK_ENABLED is defined in another place.
#define ABTI_CHECK_TRUE_RET(cond,val)                       \
    do {                                                    \
        if (ABT_CONFIG_IS_ERROR_CHECK_ENABLED && !(cond)) { \
            return (val);                                   \
        }                                                   \
    } while(0)
```
Note that such a compile-time constant is propagated at compile time and thus this value
is not evaluated at runtime, so there is no performance penalty.

There are two additional merits:
- Fewer `#ifndef` - `#endif`. Code is cleaner.
- Refer to all arguments and labels even if the error check feature is disabled. This does not affect the runtime behavior but can suppress some compiler's "unused" warnings.

### 2. Type check in `ABTI_CHECK-xxx` functions.

The comparison with `NULL` does not check the type of the pointer, so the given argument type might be different from what is intended. This PR casts a `NULL` pointer to a specific type so that it causes a warning or an error if a given pointer has a different type.

For example,
```c
    ABTI_future *p_future = ABTI_future_get_ptr(future);
    ABTI_CHECK_NULL_EVENTUAL_PTR(p_future); // should be FUTURE_PTR,
```
The above code was compiled without any warning since p_future is compared with `NULL` in `ABTI_CHECK_NULL_EVENTUAL_PTR` (and if `NULL` it returns `ABT_ERR_INV_EVENTUAL`, which is wrong).
The new macro casts `NULL` to `(ABTI_eventual *)NULL`, so using a wrong `ABTI_CHECK_xxx` becomes a warning.
```
warning: comparison of distinct pointer types lacks a cast
```

### DIFF
--- a/src/futures.c
+++ b/src/futures.c
@@ -319,7 +319,7 @@ int ABT_future_reset(ABT_future future)
 {
     int abt_errno = ABT_SUCCESS;
     ABTI_future *p_future = ABTI_future_get_ptr(future);
-    ABTI_CHECK_NULL_EVENTUAL_PTR(p_future);
+    ABTI_CHECK_NULL_FUTURE_PTR(p_future);
 
     ABTI_spinlock_acquire(&p_future->lock);
     p_future->ready = ABT_FALSE;
--- a/src/include/abti_error.h
+++ b/src/include/abti_error.h
@@ -190,6 +190,15 @@
         }                                                               \
     } while (0)
 
+#define ABTI_CHECK_NULL_XSTREAM_BARRIER_PTR(p)      \
+    do {                                            \
+        if (ABTI_IS_ERROR_CHECK_ENABLED             \
+            && p == (ABTI_xstream_barrier *)NULL) { \
+            abt_errno = ABT_ERR_INV_BARRIER;        \
+            goto fn_fail;                           \
+        }                                           \
+    } while (0)
+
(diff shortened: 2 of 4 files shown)


shintaro-iwasaki:
test:jenkins/all

PR #135, opened 2019-12-10
xstream: start ESes in ABT_xstream_create_basic()

### LINKED ISSUE #133: Inconsistent behaviour among ABT_xstream_create_xxx
The current behaviour of `ABT_xstream_create_xxx` is incosistent among them:
* `ABT_xstream_create` and `ABT_xstream_create_with_rank` create an ES and start it.
* `ABT_xstream_create_basic` creates an ES, but does not start it.

I don't know what is the expected behaviour, but the previous behaviour before commit 99796faf60610f510ee5719e0dd82ea8c65a0e73 was to actually start ESes in `ABT_xstream_create_basic`. This was caused by simple replacement of `ABT_xstream_create` with `ABTI_xstream_create`.

I think separating `ABT_xstream_create` and `ABT_xstream_start` is more explicit and better, although it will be a breaking change. What is your opinion?

s417-lama:
Fixed the change in 99796faf60610f510ee5719e0dd82ea8c65a0e73.
ABT_xstream_create_basic() should start the created ES immediately as
well as ABT_xstream_create().

Fixes #133.

### DIFF
--- a/src/stream.c
+++ b/src/stream.c
@@ -159,6 +159,10 @@ int ABT_xstream_create_basic(ABT_sched_predef predef, int num_pools,
     abt_errno = ABTI_xstream_create(&p_local, p_sched, &p_newxstream);
     ABTI_CHECK_ERROR(abt_errno);
 
+    /* Start this ES */
+    abt_errno = ABTI_xstream_start(p_local, p_newxstream);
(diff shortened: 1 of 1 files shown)


shintaro-iwasaki:
test:jenkins/all

PR #136, opened 2020-01-27
global: restore cpuset (affinity) at ABT_finalize()

s417-lama:
## Problem

The `cpuset` (affinity) of the main pthread is changed if `ABT_ENV_SET_AFFINITY` is on, but it is not restored at finialization, which limits the number of cores in the following execution.

For example,

```c
/* 1st exec */
ABT_init(0, NULL);
...
ABT_finalize();

/* 2nd exec */
ABT_init(0, NULL);
...
ABT_finalize();
```

the above second execution only uses one core even if multiple ESes are created, since the `cpuset` of the main pthread is set to `core 0` at `ABT_init()`.

## Solution

Save `cpuset` at `ABT_init()`, and restore it at `ABT_finalize()`.

### DIFF
--- a/src/arch/abtd_affinity.c
+++ b/src/arch/abtd_affinity.c
@@ -39,6 +39,7 @@ enum {
 };
 static int g_affinity_type = ABTI_ES_AFFINITY_DEFAULT;
 static cpu_set_t g_cpusets[CPU_SETSIZE];
+static cpu_set_t g_initial_cpuset;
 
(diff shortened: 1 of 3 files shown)


shintaro-iwasaki:
Thank you for the bug report and making a patch! It looks good to me. If it passes tests, I will merge it.

shintaro-iwasaki:
test:jenkins

shintaro-iwasaki:
test:jenkins/all

PR #137, opened 2020-02-06
configure.ac: fix __atomic detection code

shintaro-iwasaki:
Using an assignment statement (i.e., `weak = 0`) as a function argument incurs a compilation error with `-Werror`, preventing detection of `__atomic` builtins in configure.
Specifically, the error comes from the following:
```c
// GCC-9.2.0
int new = 0, old = 0, val = 0, weak;
val = __atomic_compare_exchange(&val, &old, &new, weak = 0,
                                __ATOMIC_ACQ_REL, __ATOMIC_ACQUIRE);
```
causes the following warning, which becomes an error with `-Werror`.
```
warning: suggest parentheses around assignment used as truth value [-Wparentheses]
```
This patch fixes it by not using `(weak = 0)`; there's little sense to label the argument anyway.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -622,13 +622,13 @@ AS_IF([test "x$enable_dynamic_promotion" = "xyes" -a "x$enable_fcontext" != "xno
 # check if __atomic builtins are supported
 AC_TRY_COMPILE(
 [#include <stdint.h>],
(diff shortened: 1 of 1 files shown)


shintaro-iwasaki:
test:jenkins/all

shintaro-iwasaki:
test:jenkins/all

Issue #17, opened 2017-04-07
List of features to deprecate


halimamer:
There are a number of features that were originally added to the public Argobots API that proved unnecessary or break some intentions. This issue keeps track of these features that will be considered for deprecation in the future.

1.  User-provided stacks (c.f. https://github.com/pmodels/argobots/issues/16)
2. Allowing busy-waiting, i.e., blocking, inside a tasklet (c.f. https://github.com/pmodels/argobots/issues/15)
3. The event interface

carns:
FWIW, the Mochi project doesn't use either of these features; we wouldn't notice them being removed :)

shintaro-iwasaki:
`ABT_xstream_create_with_rank`:
it does not seem very necessary. It just makes the ES management complicated.

`ABT_thread_retain`/`ABT_thread_release`/`ABT_task_retain`/`ABT_task_release`:
1. They are totally broken now.
2. There's no good fix without hurting performance (these features force atomic operations (just for counters) to the runtime).

`ABT_thread_set_arg`/`ABT_thread_get_arg`/`ABT_task_get_arg`/`ABT_self_set_arg`/`ABT_self_get_arg`: 
1. The runtime cannot guarantee when the arguments  are read by ULTs.
2. Currently this functionality is used to embed user data (a pointer) to a ULT, but again, it is not the intended use. We should create `ABT_thread_set_data` instead.

pavanbalaji:
@shintaro-iwasaki It might be useful to create a PR removing these features and have @carns and others (e.g., folks from the DAOS project) approve it.

carns:
FWIW we don't use any of those specific functions, but the PR approach seems reasonable to me.

shintaro-iwasaki:
The information is obsolete, so I will close this issue. The following is the summary.
 
1. [User-provided stacks] Argobots continues to support it.
 It is useful in some cases (#274)

2. [Allowing busy-waiting inside a tasklet] Argobots continues to support it. Argobots 2.0 allows more.
 All blocking operations support tasklets from Argobots 2.0. It is the user's responsibility to avoid a deadlock.

3. [The event interface] Argobots 1.0 has removed it
 It was unused.

4. [`ABT_xstream_create_with_rank()`] Argobots continues to support it.
  It's mostly useless but we cannot easily remove it. `ABT_xstream_set_rank()` should be also problematic, but at least we can keep supporting it without any performance overheads.

[...]

PR #138, opened 2020-02-06
Deprecate ABT_{thread/task}_{retain/release}

shintaro-iwasaki:
As discussed in #17, `ABT_thread_retain()`, `ABT_thread_release()`, `ABT_task_retain()`, and `ABT_task_release()`, which have been designed to control refcounts of threads and tasks, have the following issues:
1. Currently, they are broken (i.e., `ABT_thread_release()` does not free a thread even if the counter becomes 0).
2. Fixing it is costly since this adds atomic refcount operations on freeing threads (i.e., `ABT_thread_free()`).
3. These functions are not used in our understanding. At least, they do not appear in our tests.

This PR deprecates these functions. If there's no complaint, they will be removed in the future release.

Note that this patch does not break the current release/retain mechanism; it will just warn programmers:
```
# GCC 5.4
warning: ‘ABT_thread_retain’ is deprecated [-Wdeprecated-declarations]
    ABT_thread_retain(thread);
    ^
```

Note: `ax_gcc_func_attribute.m4` does not have a license issue and is used in MPICH, for example.
https://github.com/pmodels/mpich/blob/master/confdb/ax_gcc_func_attribute.m4

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -279,6 +279,15 @@ AC_SUBST([fctx_arch_bin])
 AM_SUBST_NOTMAKE([fctx_arch_bin])
 # check the pointer size
 AC_CHECK_SIZEOF([void *])
+# check __attribute__((deprecated))
+AX_GCC_FUNC_ATTRIBUTE(deprecated)
+if test "x$ax_cv_have_func_attribute_deprecated" = "xyes" ; then
+  ABT_DEPRECATED="__attribute__((deprecated))"
+else
+  ABT_DEPRECATED=""
+fi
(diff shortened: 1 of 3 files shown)


shintaro-iwasaki:
test:jenkins/all

shintaro-iwasaki:
test:jenkins/all

shintaro-iwasaki:
test:jenkins/all

shintaro-iwasaki:
These functions are marked as deprecated, but they are not removed immediately; please let us know if these functions are actually used.

PR #139, opened 2020-02-06
Add a static assert macro

shintaro-iwasaki:
This PR adds a static assert macro, `ABTI_STATIC_ASSERT()`. Currently Argobots only supports `ABTI_ASSERT()`, which checks a given condition at execution time. However, some conditions are compile-time constant and thus should be checked at compile time. This macro uses a trick to implement a static assert macro similar to `BUILD_BUG_ON()` in the Linux kernel (http://lxr.linux.no/linux+v2.6.26.5/include/linux/kernel.h#L494).

### DIFF
--- a/src/include/abti_error.h
+++ b/src/include/abti_error.h
@@ -23,6 +23,11 @@
         }                                  \
     } while(0)
 
+#define ABTI_STATIC_ASSERT(cond)                \
+    do {                                        \
+        ((void)sizeof(char[2 * !!(cond) - 1])); \
+    } while(0)
+
 #define ABTI_CHECK_INITIALIZED()                                     \
     do {                                                             \
         if (ABTI_IS_ERROR_CHECK_ENABLED && gp_ABTI_global == NULL) { \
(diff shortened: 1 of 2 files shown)


shintaro-iwasaki:
test:jenkins/all

raffenet:
If you aren't worried about older compilers, you can also use `_Static_assert`. https://en.cppreference.com/w/c/language/_Static_assert

shintaro-iwasaki:
Thanks for the comment!

A compiler-supported static assert has several benefits and I would like to use it if this is supported by most C compilers.
- Its error message is more readable.
- It can be used outside a statement scope, which looks cleaner in some cases
```c
#define ABTI_STATIC_ASSERT(cond) ((void)sizeof(char[2 * !!(cond) - 1]));

_Static_assert(sizeof(int) == 4, "int is not 4 bytes"); // OK
// ABTI_STATIC_ASSERT(sizeof(int) == 4); // not allowed outside the function

void func() {
  _Static_assert(sizeof(short) ==2, "short is not 2 bytes"); // OK
  ABTI_STATIC_ASSERT(sizeof(short) == 2); // OK
}
```
As mentioned by @raffenet, the problem is that it is a C11 feature while currently Argobots does not use any C11 feature. Because of the difference of functionality (i.e., `_Static_assert` can be used where `ABTI_STATIC_ASSERT` cannot), at present I don't plan to use it by detecting compiler support at configure-time.

PR #141, opened 2020-02-06
Clean up the interface of ABTI_spinlock

shintaro-iwasaki:
This PR cleans up the interface of `ABTI_spinlock`. Specifically, `ABTI_spinlock` can be initialize statically and do not require explicit finalization, but the previous interface requires users to create it dynamically and free it before a program finishes.
This PR changes the interface of `ABTI_spinlock` as follows:

Added function and macro:
- `ABTI_spinlock_clear()`: a function that initializes lock. The term "clear" is established for this usage (for example, `std::atomic_flag::clear` and `__atomic_clear`).
- `ABTI_SPINLOCK_STATIC_INITIALIZER()`: a static initializer.

Removed functions:
- `ABTI_spinlock_create()`: it is renamed to `ABTI_spinlock_clear()` since there is no "release" operation for "clear".
- `ABTI_spinlock_free()`: it is removed since it does nothing.

This change can remove wasteful `ABTI_spinlock_free()` calls from many places and simplify `global.c` and `include/abti_valgrind.h` both of which require static spinlock initializers.

### DIFF
--- a/src/barrier.c
+++ b/src/barrier.c
@@ -31,7 +31,7 @@ int ABT_barrier_create(uint32_t num_waiters, ABT_barrier *newbarrier)
 
     p_newbarrier = (ABTI_barrier *)ABTU_malloc(sizeof(ABTI_barrier));
 
-    ABTI_spinlock_create(&p_newbarrier->lock);
+    ABTI_spinlock_clear(&p_newbarrier->lock);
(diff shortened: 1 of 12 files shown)


shintaro-iwasaki:
test:jenkins/all

shintaro-iwasaki:
test:jenkins/all

PR #142, opened 2020-02-06
info: make a stack print function more robust

shintaro-iwasaki:
`ABTI_info_check_print_all_thread_stacks()` refers to `gp_ABTI_global->num_xstreams` and accesses execution streams based on this value without taking `gp_ABTI_global->xstreams_lock`. This behavior potentially causes an error if execution streams are dynamically created/removed while traversing and printing elements in a pool.

This patch adds a lock (`gp_ABTI_global->xstreams_lock`) to avoid this situation. In addition, introducing a critical section can remove meaningless atomic load of `gp_ABTI_global->num_xstreams` (`ABTD_atomic_load_int32(&gp_ABTI_global->num_xstreams)`).

### DIFF
--- a/src/info.c
+++ b/src/info.c
@@ -490,14 +490,23 @@ void ABTI_info_check_print_all_thread_stacks(void)
         /* This ES becomes a master. */
         double start_time = ABTI_get_wtime();
         ABT_bool force_print = ABT_FALSE;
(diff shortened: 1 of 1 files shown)


shintaro-iwasaki:
test:jenkins/all

shintaro-iwasaki:
test:jenkins/all

PR #149, opened 2020-02-27
Use proper ID to distinguish ESs, ULTs, tasklets, and external threads.

shintaro-iwasaki:
### Problems

Currently, `ABTI_unit *` is used to distinguish ULTs, tasklets, and external threads, and 'ABTI_xstream *' is used to distinguish ESs and external threads. In both cases, external threads (i.e., Pthreads) do not have an entity of `ABTI_unit` and `ABTI_xstream`, so its `pthread_t` is cast to a pointer type (`ABTI_unit *` and 'ABTI_xstream *') and assigned.

For example, `p_owner` points to either`p_thread/p_task->unit_def` (real `ABT_unit`) or `pthread_t` in the following case.
```c
struct ABTI_mutex_attr {
   ABTI_unit *p_owner;
};
int ABT_mutex_lock(...) {
  // get local unit
  ABTI_unit *p_self;
  if (!lp_ABTI_local)              p_self = (ABTI_unit *)pthread_self();
  else if(lp_ABTI_local->p_thread) p_self = &p_thread->unit_def;
  else if(lp_ABTI_local->p_task)   p_self = &p_task->unit_def;
  if (p_mutex->p_owner == p_self)
    [...]; // if the caller is an owner of this mutex, increment the nest level and return.
}
```
This design has several problems.
1. The developers might think `p_owner` really points to `ABTI_unit`, while it does not in the case of an external thread.
2. This is not portable. A `pthread_t` value might be longer than the pointer size.
3. Values are not unique. There's no guarantee that a `pthread_t` value is different from values of any `&p_thread->unit_def` and `&p_task->unit_def`.

### Solution

This PR fixes them by introducing `void *`-type ID, which uses `&lp_ABTI_local` (the address of the TLS value) to identify external threads.
```c
typedef void *ABTI_unit_id;
struct ABTI_mutex_attr {
   ABTI_unit_id owner_id;
};
int ABT_mutex_lock(...) {
  // get local unit_id
  ABTI_unit_id self_id;  
  if (!lp_ABTI_local)              self_id = (ABTI_unit_id)&lp_ABTI_local;
  else if(lp_ABTI_local->p_thread) self_id = (ABTI_unit_id)&p_thread;
  else if(lp_ABTI_local->p_task)   self_id = (ABTI_unit_id)&p_task;
  if (p_mutex->owner_id == self_id)
    [...]; // if the caller is an owner of this mutex, increment the nest level and return.
}
```
This solution 1. forbids dereference of `p_owner` because `ABTI_unit_id` is an opaque pointer, 2. is portable because it relies on only pointers, and 3. guarantees uniqueness of IDs because each value points to a real memory address.

Specifically, this PR adds two new types: `ABTI_unit_id` and `ABTI_native_thread_id`. `ABTI_unit_id` is to distinguish ULTs, tasklets, and external threads, while `ABTI_native_thread_id` is designed for execution streams and external threads.

### Performance

This change does not affect performance in the critical path if producer/consumer-check is disabled. Otherwise, there might be a slight performance difference though it should be almost negligible.

### DIFF
--- a/src/include/abti.h
+++ b/src/include/abti.h
@@ -129,6 +129,10 @@ typedef struct ABTI_stack_header    ABTI_stack_header;
 typedef struct ABTI_page_header     ABTI_page_header;
 typedef struct ABTI_sp_header       ABTI_sp_header;
 #endif
+/* ID associated with native thread (e.g, Pthreads), which can distinguish
+ * execution streams and external threads */
+struct ABTI_native_thread_id_opaque;
+typedef struct ABTI_native_thread_id_opaque *ABTI_native_thread_id;
 /* ID associated with work unit (i.e., ULTs, tasklets, and external threads) */
 struct ABTI_unit_id_opaque;
 typedef struct ABTI_unit_id_opaque *ABTI_unit_id;
--- a/src/include/abti.h
+++ b/src/include/abti.h
@@ -270,10 +270,10 @@ struct ABTI_pool {
     int32_t num_scheds;      /* Number of associated schedulers */
                              /* NOTE: int32_t to check if still positive */
 #ifndef ABT_CONFIG_DISABLE_POOL_CONSUMER_CHECK
-    ABTI_xstream *consumer;  /* Associated consumer ES */
+    ABTI_native_thread_id consumer_id; /* Associated consumer ID */
 #endif
 #ifndef ABT_CONFIG_DISABLE_POOL_PRODUCER_CHECK
-    ABTI_xstream *producer;  /* Associated producer ES */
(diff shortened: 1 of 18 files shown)


shintaro-iwasaki:
test:jenkins/all

shintaro-iwasaki:
test:jenkins/all

shintaro-iwasaki:
test:jenkins/all

shintaro-iwasaki:
test:jenkins/all

shintaro-iwasaki:
test:jenkins/all

shintaro-iwasaki:
test:jenkins/all

shintaro-iwasaki:
test:jenkins/all

Issue #143, opened 2020-02-11
Provide an interface to query supported features.


shintaro-iwasaki:
Argobots provides several compilation options, some of which disable features at all. For example, users can disable ULT cancellation, tasklet cancellation, ULT migration, stackable scheduler, supporting external threads, affinity, and FPU-preservation. However, Argobots currently does not have an interface to know whether these features are enabled at execution time. We need an interface to query this information.

There are several ways to implement it (e.g., making a function per feature, creating a query function that takes a string/key ID and returns a value/void pointer, ...). We need to check the design of widely used systems (e.g., MPI and OpenMP).

shintaro-iwasaki:
OpenMP 5.0: query function per feature.
```
int omp_get_supported_active_levels(void);
int omp_get_cancellation(void);
```

MPI 3.1: query with an index.
```
int MPI_T_enum_get_info(MPI_T_enum enumtype, int *num, char *name, int *name_len);
int MPI_T_cvar_get_info(int cvar_index, char *name, int *name_len, int *verbosity, MPI_Datatype *datatype, MPI_T_enum *enumtype, char *desc, int *desc_len, int *binding, int *scope);
```

shintaro-iwasaki:
#151 addressed this issue.

PR #151, opened 2020-03-02
info: implement a function to query Argobots configuration

shintaro-iwasaki:
This PR addresses #143 .

Argobots provides no method to programatically know which features are available, though some features such as external thread supports and error checks significantly affect the behavior of Argobots. This is problematic when users cannot control the underlying Argobots implementation, for example, when the shared library is used.

To address this problem, this PR implements a query function, `ABT_info_query_config()`, which returns a configuration parameter of Argobots.


shintaro-iwasaki:
test:jenkins/all

shintaro-iwasaki:
test:jenkins/all

PR #153, opened 2020-03-13
atomic: force atomic-type variables

shintaro-iwasaki:
#### Problem

The current Argobots has a wrapper for atomic operations but not atomic variables. This makes it vague whether a specific variable should be atomically accessed or not.

```c
struct ABTI_thread {
  ...
  uint32_t request;  // Should this variable be accessed atomically?
  uint32_t refcount; // How about this variable?
  ...
};
void ABTDI_thread_terminate(...) {
  [...];
  ABTD_atomic_store_uint32(&p_thread->request, ABTI_THREAD_REQ_TERMINATE);
  [...];
}
int ABTI_thread_revive(...)
  [...];
  p_thread->request = 0;  // Can we update it without an atomic operation?
  p_thread->refcount = 1; // How about this?
  [...];
}
```

#### Solution

To ease the programming, this PR makes atomic operations only work for *atomic-type variables*. Developers need to be aware of atomicity when accessing such an atomic-type variable.

```c
struct ABTI_thread {
  ...
  ABTD_atomic_uint32 request; // This atomic-type variable should be accessed atomically.
  uint32_t refcount; // This should not be accessed with atomic operations.
  ...
};
void ABTDI_thread_terminate(...) {
  [...];
  // Explicitly write this as release store.
  ABTD_atomic_release_store_uint32(&p_thread->request, ABTI_THREAD_REQ_TERMINATE);
  [...];
}
int ABTI_thread_revive(...)
  [...];
  // Explicitly write this as relaxed store (if the developer knows it is okay).
  ABTD_atomic_relaxed_store_uint32(&p_thread->request, 0);
  p_thread->refcount = 1; // This should be a normal store.
  // The following causes a compile-time error.
  // ABTD_atomic_relaxed_store_uint32(&p_thread->refcount, 1);
  [...];
}
```

It will not hurt performance because the resulted code should be the same after compilation. Note that MPICH has a similar atomic wrapper.

### DIFF
--- a/src/arch/abtd_thread.c
+++ b/src/arch/abtd_thread.c
@@ -70,7 +70,7 @@ static inline void ABTDI_thread_terminate(ABTI_local *p_local,
         if (p_thread->p_last_xstream == p_joiner->p_last_xstream) {
             /* Only when the current ULT is on the same ES as p_joiner's,
              * we can jump to the joiner ULT. */
-            ABTD_atomic_store_uint32((uint32_t *)&p_thread->state,
+            ABTD_atomic_store_int((int *)&p_thread->state,
                                      ABT_THREAD_STATE_TERMINATED);
             LOG_EVENT("[U%" PRIu64 ":E%d] terminated\n",
                       ABTI_thread_get_id(p_thread),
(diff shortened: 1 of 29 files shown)


shintaro-iwasaki:
test:jenkins/all

shintaro-iwasaki:
test:jenkins/all

shintaro-iwasaki:
test:jenkins/all

shintaro-iwasaki:
test:jenkins/all

shintaro-iwasaki:
test:jenkins/all

PR #158, opened 2020-03-18
unit: use is_in_pool instead of pool in ABTI_unit

shintaro-iwasaki:
### Problem

Information whether a unit is in a pool or not can be accessed by multiple threads.

https://github.com/pmodels/argobots/blob/master/src/thread.c#L1918
https://github.com/pmodels/argobots/blob/master/src/thread.c#L2321-L2327 

An `ABT_pool`-type variable, `pool`, is used to get this information, but this `pool` is accessed and updated with non-atomic operations, so it can potentially  cause a bug if multithreaded.

### Solution and benefits

This PR simplifies the overall `is_in_pool` mechanism by introducing an `ABT_atomic_int`-type variable, `is_in_pool`. Good things are as follows:
- Atomic acquire/release operations can guarantee completion of pool operations.
- This change reduces the size of `ABTI_unit` by 8 bytes.
  - This is demanded to add a new `void *` member variable to implement `ABT_{thread/task}_{get/set}_arg()`.

#### Note

Note that a parent-pool check is removed from `pool_remove` operations, but this is user's responsibility and thus not required.

### DIFF
--- a/src/cond.c
+++ b/src/cond.c
@@ -189,9 +189,9 @@ int ABT_cond_timedwait(ABT_cond cond, ABT_mutex mutex,
     ABTD_atomic_int32 ext_signal = ABTD_ATOMIC_INT32_STATIC_INITIALIZER(0);
 
     p_unit = (ABTI_unit *)ABTU_calloc(1, sizeof(ABTI_unit));
-    /* Check size if ext_signal can be stored in p_unit->pool. */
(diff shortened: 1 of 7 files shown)


shintaro-iwasaki:
test:jenkins/all

shintaro-iwasaki:
test:jenkins/all

shintaro-iwasaki:
test:jenkins/all

PR #171, opened 2020-05-13
abt.h: remove unnecessary const

shintaro-iwasaki:
`ABT_xstream_set_rank()` was declared as follows, but the const qualifier to a scalar-type argument is meaningless.
```c
int ABT_xstream_set_rank(ABT_xstream xstream, const int rank);
```
This change keeps the same function type, so it won't break the compatibility; for example, the following is allowed (without any compiler warning) as far as I checked.
```c
int ABT_xstream_set_rank(ABT_xstream xstream, int rank); // no const
void func() {
  typedef void (*old_fn_type)(ABT_xstream, const int); // const here
  old_fn_type fn_type = ABT_xstream_set_rank; // no warning
}
```


shintaro-iwasaki:
test:argobots/all

shintaro-iwasaki:
Please let us know if it should break your program.

PR #172, opened 2020-05-13
abt.h: add meaningful const

shintaro-iwasaki:
## Problem

Some Argobots functions take non-const pointer-type arguments that are actually read-only, which confuses users and, in the worst case, requires `const_cast` or C-style cast. For example,

```c
// Current declaration
int ABT_xstream_set_affinity(ABT_xstream xstream, int cpuset_size, int *cpuset);

void func1(int cpuset_size, const int *cpuset) {
	// Compilation error/warning. Explicit cast is needed.
	// ABT_xstream_set_affinity(g_xstream, cpuset_size, cpuset);
}
void func2(const std::vector<int> &cpuset) {
	// Compilation error/warning.  Explicit cast is needed.
	// ABT_xstream_set_affinity(g_xstream, cpuset.size(), &cpuset[0]);
}
```

## Solution

This patch adds `const` if values are unchanged.

## Expected impact

Although it does not affect most user-programs, strictly speaking, this patch breaks the compatibility unlike #171.

```c
typedef int set_affinity_fn(ABT_xstream, int, int*);
int ABT_xstream_set_affinity(ABT_xstream xstream, int cpuset_size, int *cpuset);
int New_xstream_set_affinity(ABT_xstream xstream, int cpuset_size, const int *cpuset);

set_affinity_fn set_affinity1 = ABT_xstream_set_affinity; // Okay
set_affinity_fn set_affinity2 = New_xstream_set_affinity; // Error/warning
```

This PR should be merged after we understand the set of functions of the next Argobots release(s).

### Note

It might be a little bit confusing, but all `ABT_xxx` (except `ABT_sched_def` and `ABT_pool_def`) are scalar-type handlers, so `const ABT_xxx` does not mean a pointed object is unchanged.

```c
void func(const ABT_thread thread) {
	ABT_thread my_thread = thread; // It is okay since it is just a scalar-value copy
	// thread = ABT_THREAD_NULL; // This is not allowed because of "const".
	// Of course, now my_thread can be passed to any Argobots functions.
}
```

It means that, for example, the following means that each `ABT_pool`'s entity can be changed (e.g., by pushing a thread to that pool), but the content of `pool_list` is unchanged before and after the function.

```c
int ABT_thread_create_many(int num, ABT_pool *pool_list, [...]);
int New_thread_create_many(int num, const ABT_pool *pool_list, [...]);

void func() {
	ABT_pool pools[] = {g_pool0, g_pool1};

	New_thread_create_many(2, pools, [...]);
	assert(pools[0] == g_pool0 && pools[1] == g_pool1); // A function type tells yes.

	ABT_thread_create_many(2, pools, [...]);
	assert(pools[0] == g_pool0 && pools[1] == g_pool1); // Need to read a document to see if it always holds.
}
```


shintaro-iwasaki:
When we change the API, we should also 
- remove `uint32/64_t` from the header (it is not a "must" type in C)
- rethink signedness of integer arguments
- make a function that uses `timespec` optional (exclude it if the environment does not have `timespec_t`).

shintaro-iwasaki:
This PR is stale. Let's close this.

PR #175, opened 2020-05-15
Robust logging

shintaro-iwasaki:
Although `ABTI_log` functions are for debugging, they do not work because of SEGV when  `lp_local == NULL` and `lp_ABTI_log == NULL`. The log interface should be robust so that it can be used everywhere. This PR fixes it.

Note that there is no performance impact if a user does not explicitly set `--enable-debug=log` or stronger; even with this option, there should not be performance difference.


shintaro-iwasaki:
test:argobots/all

shintaro-iwasaki:
test:argobots/all

PR #176, opened 2020-05-16
Remove ABTI_local

shintaro-iwasaki:
This PR is for code refactoring. Argobots does not change Pthreads-xstream mapping, so there is one-to-one mapping between `ABTI_local` and `ABTI_xstream`, but they are managed independently.

This PR unifies `ABTI_local` and `ABTI_xstream`, which can remove some dereferences. The code becomes less confusing.

Basically there should be almost no performance difference (or even slightly better). No new functionality is added.


shintaro-iwasaki:
test:argobots/all

shintaro-iwasaki:
test:argobots/all

shintaro-iwasaki:
test:argobots/all

PR #177, opened 2020-05-16
Miscellaneous thread allocation fixes

shintaro-iwasaki:
This PR fixes small issues and cleans up `ABTI_mem_alloc_thread()`.

#### 1. A (potential) bug when `ABT_thread_attr_set_stack()` is called twice

The following code does not reset a "stack" variable in `ABTI_thread_attr`. This PR fixes it.
```c
ABT_thread_attr_set_stack(attr, stack, stacksize1);
ABT_thread_attr_set_stack(attr, NULL, stacksize2);
// Since NULL is set here, p_stack should be internally updated
```

#### 2. SEGV when `ABT_thread_create_XXX` is called with the default thread attribute on an external thread.

The following code tries to access a local memory pool, which causes SEGV. This PR fixes it.
```c
external_thread():
  ABT_thread_attr attr;
  ABT_thread_attr_create(&attr); // attr is initialized.
  ABT_thread_create(..., attr, ...);
```

#### 3. Remove a comparison

This slightly simplifies the complicated branch in `ABTI_mem_alloc_thread()`. This also compensates the additional branch added by 2.

They are not performance critical, so the performance difference is not visible.

### DIFF
--- a/src/include/abti_mem.h
+++ b/src/include/abti_mem.h
@@ -173,7 +173,7 @@ static inline ABTI_thread *ABTI_mem_alloc_thread(ABTI_xstream *p_local_xstream,
         }
 
         stacksize = p_attr->stacksize;
-        if (stacksize != def_stacksize || stacktype == ABTI_STACK_TYPE_MALLOC) {
+        if (stacktype == ABTI_STACK_TYPE_MALLOC) {
             /* Since the stack size requested is not the same as default one,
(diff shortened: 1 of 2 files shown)


shintaro-iwasaki:
test:argobots/all

shintaro-iwasaki:
test:argobots/all

PR #179, opened 2020-05-22
Memory code cleanup

shintaro-iwasaki:
This PR applies code refactoring to `malloc.c` and `abti_mem.h`.

1. Create `ABTU_xxx_largepage()` functions and use them in `malloc.c`.
    - `malloc.c` does not need to handle detailed `mmap` options, which are now handled in `util/largepage.c`.
2. Clean up control flow in `ABTI_mem_alloc_thread()`
    - The current code is too complicated to understand which path is finally taken with a certain situation (e.g., default vs explicit MEMPOOL under external threads). This PR fixes it, though it makes the code a bit longer.

The logic will not be changed by this PR.


shintaro-iwasaki:
test:argobots/all

Issue #170, opened 2020-05-05
Clarify "execution contexts" in Argobots


shintaro-iwasaki:
Argobots has quite a few implicit execution contexts that have different programming restrictions and change behavior of functions: ULT context, tasklet context, external-thread context, scheduler context, and pool context (and some call-back contexts).  This introduces several unknown and unreasonable restrictions.  For example, in the case of `ABT_thread_yield()`:

- `ABT_thread_yield()` in a ULT context works as expected.
- `ABT_thread_yield()` in a tasklet is an "error".
- `ABT_thread_yield()` in a ULT-based scheduler context does not work (broken).
- `ABT_thread_yield()` in a user-defined pool might work in a few cases but is basically forbidden.
- `ABT_thread_yield()` in a external-thread context is "ignored" (returns `ABT_SUCCESS`).

We should clarify them in a document, but first, we should make the current situation more reasonable by relaxing some restrictions.

shintaro-iwasaki:
Although it is still not clear which function can be called in a callback function, all the functions are examined by #232. Basically there are only three contexts (except for a few functions): ULT/tasklet/external threads, ULT/tasklet, and only ULT. For example, `ABT_thread_yield()` is the third type: it can be called only on a ULT.

I believe this classification is sufficient now. Let's reopen this issue if this becomes a problem again.

PR #181, opened 2020-05-27
sched: narrow the gap between scheduler and thread

shintaro-iwasaki:
## Problem

In Argobots, a scheduler (`ABTI_sched`) is treated like a first-class schedulable work unit though its underlying implementation is either a ULT or a tasklet. This work unit type creates a weird context called *a scheduler context*, which is visible to users in a user-defined scheduler. As mentioned in #170, this scheduler context is restricted by the underlying implementation (i.e., ULT or tasklet), but also has its own semantics (e.g., `ABT_self_xxx()` is disallowed). In the Argobots runtime, this scheduler handling is a mess (e.g., `ABTI_LOG_SET_SCHED()`).  Since essentially a scheduler is (and should be) a ULT in terms of threading capabilities, we should narrow this gap as much as possible.

## Solution

This PR does not fix all the issues, but it removes most part of special handling for scheduler contexts in Argobots.  Now a stackable scheduler can be simply invoked as a thread.

## Known impact

This PR adds a small overhead on context switching since it adds flag updates that are not performed in scheduler contexts. This flag management is unavoidable to unify the contexts (especially in order to treat a scheduler like a normal ULT).

This does not break the current API.

### DIFF
--- a/src/arch/abtd_thread.c
+++ b/src/arch/abtd_thread.c
@@ -5,12 +5,10 @@
 
 #include "abti.h"
 
-static inline void ABTD_thread_terminate_thread(ABTI_xstream *p_local_xstream,
-                                                ABTI_thread *p_thread);
-static inline void ABTD_thread_terminate_sched(ABTI_xstream *p_local_xstream,
-                                               ABTI_thread *p_thread);
+static inline void ABTD_thread_terminate(ABTI_xstream *p_local_xstream,
+                                         ABTI_thread *p_thread);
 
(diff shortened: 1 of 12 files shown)


shintaro-iwasaki:
test:argobots/all

shintaro-iwasaki:
test:argobots/all

shintaro-iwasaki:
test:argobots/all

shintaro-iwasaki:
test:argobots/all

PR #182, opened 2020-05-27
sched: use a ULT for a tasklet-type stackable scheduler

shintaro-iwasaki:
The current implementation allows a tasklet-type stackable scheduler, which is very restrictive in terms of functionality though it can reduce only little overheads since basically schedulers are not invoked frequently. 

This PR modifies a tasklet-type stackable scheduler to use a ULT internally, while it keeps the external interface that allows users to specify a tasklet type (i.e., Argobots still accepts `ABT_SCHED_TYPE_TASK`). It can simplify the implementation in Argobots.

Note that this does not break the existing API and its semantics since previously there was no way for users to get an underlying work unit implementation in a scheduler context.


shintaro-iwasaki:
test:argobots/all

Issue #178, opened 2020-05-20
Naive lock-free stack in mempool causes an ABA problem


shintaro-iwasaki:
Argobots mempool uses a naive lock-free stack (which you can easily find as an example of bad lock-free stack) to manage memory pools for `ABTI_thread` and `ABTI_task`. This implementation does not take care of the ABA problem (https://en.wikipedia.org/wiki/ABA_problem) and, although it is "rare", our usage of memory pool does not really avoid this issue.

Specifically, it can happen as follows.
1. While one execution stream is taking a descriptor of either `ABTI_thread` and `ABTI_task`
    - This element is called `A`. Here we assume that `A->next` is `B`,
2. another execution streams takes `A`,
3. other execution streams modify the stack, for example, takes `B`, and
4. that descriptor `A` is returned to a pool but at this point `A->next` is `C`.
5. The execution stream of 1. performs CAS while it is still assuming `A->next` is `B`.
    - This succeeds since the stack top is actually `A`. The stack top becomes `B` (= `A->next`)
    - Note that `B` has been already removed from this stack in 3.
6. Another execution stream takes `B` from this stack.

This may cause SEGV because this `B` might be being used somewhere else. In reality, Argobots uses pretty large local cache and it is "rare" if 2 - 4 occur during 1, but this can cause a really hard-to-find bug.

This must be fixed immediately.

shintaro-iwasaki:
The easiest way is adding a lock, which loses lock-freedom and potentially performance. Anyway the current memory pool code is extremely messy (two memory pool implementations exist for `ABTI_thread` and `ABTI_task`, which behave slightly differently for no valid reason and these variable/struct names are extremely confusing), so they should be fixed too.

PR #183, opened 2020-05-29
mempool: lock-free bucket-based memory pools for threads and tasks 

shintaro-iwasaki:
## Problems

The current memory pools for threads and tasks use the following algorithms:

### Thread pool

- [Return] The thread pool returns an individual element to the global pool *one by one* if the local pool is full (i.e., more than `ABT_MEM_MAX_NUM_STACKS`).
- [Allocate] The thread pool takes all elements from a global pool if the local pool is empty.
    - It can take more than `ABT_MEM_MAX_NUM_STACKS` of elements.

### Task pool

- [Return] The task pool returns an individual element if that task is not created on this execution stream.
  - It caches locally-created tasks in the local pool (no upper limit).
- [Allocate] The task pool checks all the pages locally created by this execution stream and if no free tasks are available (e.g., returned in the [Return] part), allocate a new page and create headers.
  - Increase of the number of pages just increase the allocation time.

Both have several issues:
- The memory pool for threads and one for tasks use totally different algorithms without any specific reason, so developers need to maintain two different algorithms.
- Both pool algorithms do not a maximum number of elements in a local cache, so it requires significant memory footprint.
   - For example, even if the maximum number of threads/tasks at any given point is "N", the consumed memory can be "N  * # of ESs" in the worst case (since there is no upper limit).
- Both algorithms use a naive stack implementation, which has the ABA issue (#178)
- The interface is not very generic, so it needs another hard-coded memory pool for other descriptors (if needed).

## Solution

This PR creates a generic lock-free bucket-based memory pool that allows users to set the strict capacity of every local pool. The new implementation has the following merits:

- The ABA issue is solved in a lock-free manner by using a pointer+tag parallel LIFO algorithm (https://en.wikipedia.org/wiki/ABA_problem#Tagged_state_reference) if the architecture supports 16-byte CAS.
   - As far as I tried, the following environments are supported:
      - Typical Intel x86/64 with ICC17+, GCC4.8+, and Clang3.9+
      - POWER8/9 with XLC16 and GCC4.8+. No Clang (including a pretty new one) since it does not recognize 128-bit LL/SC inline assembly.
      - 64-bit ARM with GCC4.8+. No Clang (including a pretty new one) since it does not recognize 128-bit LL/SC inline assembly.
   - If not supported, it falls back to a spinlock-based implementation (and thus not lock-free).
- All the data transfer between a local pool and a global pool is per-bucket (multiple elements rather than a single element), which can reduce the overheads
   - This is also lock-free.
- Each local pool has a stricter upper limit of the capacity, so if the program uses at most "N" elements in the program, Argobots consumes at most "N + local-pool capacity * # of ESs".
   - This guarantee is not very strict because per-page memory allocation is performed in a lock-free manner (i.e., not serialized), so the total memory consumption can be a little bit exceeded in the worst case.
- Both use the same pool implementation, so developers need to optimize only one implementation if necessary.
- The interface is generic, so it is easy to create another memory pool.

## Issues

The new algorithm is **not always faster than the existing implementations**. Because of the difference of memory access order and cache access patterns, I observed 60% slowdown (and up to 2500% speedup) with fork-join mircobenchmarks. I arbitrarily set the local pool capacity, but this tuning can also negatively affect the performance in some cases. For now, because of the ABA problem (#178), I do not think the current implementation is better than this, but if the application performance is noticeably changed by this PR, please let me/us know so that we can fix it.


shintaro-iwasaki:
test:argobots/all

shintaro-iwasaki:
test:argobots/all

src/include/asm/Makefile.mk
```diff
@@ -0,0 +1,9 @@
+# -*- Mode: Makefile; -*-
+#
+# See COPYRIGHT in top-level directory.
+#
+
+if ABT_USE_INT128_ATOMIC
+noinst_HEADERS = \
+	include/asn/abtd_asm_int128_cas.h
```
shintaro-iwasaki, line 8: asm

src/mem/mem_pool.c
```diff
@@ -0,0 +1,260 @@
...
+}
+
+static inline ABTI_mem_pool_header *
+ABTI_mem_pool_lifo_elem_to_header(ABTI_sync_lifo_element *lifo_elem)
+{
+    return (
+        ABTI_mem_pool_header *)(((char *)lifo_elem) -
+                                offsetof(ABTI_mem_pool_header, bucket_info) +
+                                offsetof(ABTI_mem_pool_header_bucket_info,
+                                         lifo_elem));
```
shintaro-iwasaki, line 23: (offsetof(ABTI_mem_pool_header, bucket_info) + offsetof(ABTI_mem_pool_header_bucket_info, lifo_elem))

src/mem/mem_pool.c
```diff
@@ -0,0 +1,260 @@
...
+    size_t header_size, size_t page_size,
+    const ABTU_MEM_LARGEPAGE_TYPE *lp_type_requests, int num_lp_type_requests,
+    size_t alignment_hint)
+{
+    p_global_pool->num_headers_per_bucket = num_headers_per_bucket;
+    p_global_pool->header_size = header_size;
+    p_global_pool->page_size = page_size;
+
+    /* lp_type_requests is a constant-sized array, so num_lp_type_requests
+     * should be smaller than that value. */
```
shintaro-iwasaki, line 37: Size check

src/mem/mem_pool.c
```diff
@@ -0,0 +1,260 @@
...
+    ABTD_atomic_relaxed_store_ptr(&p_global_pool->p_mem_page_empty, NULL);
+    ABTI_sync_lifo_init(&p_global_pool->bucket_lifo);
+    ABTI_spinlock_clear(&p_global_pool->partial_bucket_lock);
+    p_global_pool->partial_bucket = NULL;
+}
+
+void ABTI_mem_pool_destroy_global_pool(ABTI_mem_pool_global_pool *p_global_pool)
+{
+    /* All local pools must be released in advance.
+     * Because all headers are from memory pages, need to free only pages. */
```
shintaro-iwasaki, line 56: each header does not need to be freed.

src/mem/mem_pool.c
```diff
@@ -0,0 +1,260 @@
...
+    }
+    const int num_headers_per_bucket = p_local_pool->num_headers_per_bucket;
+    ABTI_mem_pool_header *cur_bucket = p_local_pool->buckets[bucket_index];
+    if (cur_bucket->bucket_info.num_headers == num_headers_per_bucket) {
+        /* The last bucket is also full. Return the last bucket as well. */
+        ABTI_mem_pool_return_bucket(p_local_pool->p_global_pool,
+                                    p_local_pool->buckets[bucket_index]);
+    } else {
+        ABTI_mem_pool_global_pool *p_global_pool = p_local_pool->p_global_pool;
+        /* Return headers in the last bucket to bucket_index. */
```
shintaro-iwasaki, line 104: to bucket index ...?

shintaro-iwasaki:
test:argobots/all

shintaro-iwasaki:
test:argobots/all

shintaro-iwasaki:
test:argobots/all

shintaro-iwasaki:
test:argobots/all

shintaro-iwasaki:
Finally I could confirm that this PR works on POWER 9 and 64-bit ARM in addition to x86/64 with various compilers including GCC 4.8, 6.5, 8.3, 9.2, Clang 3.9, 7.0, 9.0, 10.0, and architecture-specific compilers such as ICC 18, 19, 20 and XLC 16.

Note that this change might cause some performance issues; please tell us if you encounter any performance issue.

PR #186, opened 2020-06-04
log: replace LOG_EVENT() with LOG_DEBUG() 

shintaro-iwasaki:
## Problem

Argobots has two logging interface: `LOG_EVENT()` and `LOG_DEBUG()`.

- `LOG_EVENT()` prints an event message. It prints the information of the underlying execution stream and ULT/tasklet in addition to the event message.
- `LOG_DEBUG()` prints a debug message. It prints the information of `__FILE__` and `__LINE__` in addition to the event message.

Those two are very similar and both are used primarily for debugging (since these logging interface is quite heavy). However, they are controlled by different environmental variables (`ABT_USE_LOG` and `ABT_USE_DEBUG`), which is confusing.

## Solution

This PR removes `LOG_EVENT()` and uses `LOG_DEBUG()` instead. The implementation of `LOG_DEBUG()` is actually replaced by that of `LOG_EVENT()`, however, since `LOG_EVENT()` prints more useful information. The log mode can be switched by `ABT_USE_LOG=0|1`.

## Impact

With the default setting, no impact. When logging is enabled (`./configure --enable-debug=XXX`, which is turned off by default), the printed messages might get slightly different, but it will not affect the most messages.

Note that this PR neither degrades nor improves the contents of debug messages.

### Note

This PR contains some minor cleanup.

### DIFF
--- a/src/arch/abtd_thread.c
+++ b/src/arch/abtd_thread.c
@@ -47,7 +47,7 @@ static inline void ABTD_thread_terminate(ABTI_xstream *p_local_xstream,
              * we can jump to the joiner ULT. */
             ABTD_atomic_release_store_int(&p_thread->state,
                                           ABT_THREAD_STATE_TERMINATED);
-            LOG_EVENT("[U%" PRIu64 ":E%d] terminated\n",
+            LOG_DEBUG("[U%" PRIu64 ":E%d] terminated\n",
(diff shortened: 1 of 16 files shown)


shintaro-iwasaki:
test:argobots/all

PR #187, opened 2020-06-05
fcontext: fix ppc64 assembly code

shintaro-iwasaki:
`make_fcontext()` did not save `FPSCR` and `VSCR`, so threads resumed by `jump_fcontext()` or `take_fcontext()` restored invalid `FPSCR` and `VSCR`. This sometimes makes results of floating operations wrong.

I saw this issue in a vectorized program (with VSX) on POWER 9, but it may happen in a non-vectorized program on other POWER architectures.

This PR fixes it.

### DIFF
--- a/src/arch/fcontext/make_ppc64_sysv_elf_gas.S
+++ b/src/arch/fcontext/make_ppc64_sysv_elf_gas.S
@@ -130,6 +130,8 @@
  *                                                     *
  *******************************************************/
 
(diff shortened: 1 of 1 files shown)


shintaro-iwasaki:
This fix cannot be checked by Jenkins since it only has x86/64 nodes, so I do not run a Jenkins test here. Note that I have checked that that this fix worked on POWER 9.

PR #188, opened 2020-06-08
threads/tasks: create a work unit abstraction in Argobots

shintaro-iwasaki:
## Problems

Threads (ULTs) and tasks (tasklets) are schedulable work units, but they have totally different implementations in Argobots.  Some are necessary because of the difference of features (e.g., only threads can yield), but more than 50% is common.  However, Argobots lacks this abstraction, so the common part is implemented different by creating two functions (often as `xxx_task()` and `xxx_thread()`).  This makes code maintenance challenging.

## Solutions

As a first step, this PR moves most member variables in `ABTI_thread` and `ABTI_task` to `ABTI_unit` to work it as the work unit abstraction.  For this unification, several constant variables are changed to share the same values between threads and tasks.

## Impact

Although this PR changes a lot of code, the performance should not be (visibly) changed. There would be a small difference because this PR reorders member variables of `ABTI_thread` and `ABTI_task`, but the number of dereferences is not increased; this PR slightly reduces indeed.  This PR keeps API/ABI.

### Note

This PR does not solve all the problems mentioned above.  The following PR should use `ABTI_unit` to remove the common part that remains in Argobots.

The last ~two~ three commits clean up code regarding `ABTI_unit` and `ABTI_thread`.


shintaro-iwasaki:
test:argobots/all

shintaro-iwasaki:
test:argobots/all

shintaro-iwasaki:
test:argobots/all

PR #189, opened 2020-06-08
mem: fix memory allocation code when the memory pool is disabled

shintaro-iwasaki:
`--disable-mem-pool` disables the memory pool optimizations in Argobots, but the current Argobots cannot be compiled with this option since it has been overlooked for a while. This PR fixes this issue.

 `--disable-mem-pool` is added to Jenkins tests (as one of the random configurations), so this option will be tested regularly.

Note that this option is useful if one uses an external scalable memory allocator (e.g., tcmalloc and jemalloc).

### DIFF
--- a/src/include/abti_mem.h
+++ b/src/include/abti_mem.h
@@ -8,7 +8,6 @@
 
 /* Memory allocation */
 
-#ifdef ABT_CONFIG_USE_MEM_POOL
 enum {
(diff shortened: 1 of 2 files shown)


shintaro-iwasaki:
test:argobots/all

shintaro-iwasaki:
test:argobots/all

shintaro-iwasaki:
test:argobots/all

PR #191, opened 2020-06-08
Cleanup thread migration code

shintaro-iwasaki:
This PR simplifies the current thread migration code, which is complicated and costly.

This PR does not change the current API/ABI. This change should improve the performance if the migration feature is used; otherwise there should be no visible performance difference.


shintaro-iwasaki:
test:argobots/all

PR #192, opened 2020-06-09
Simplify thread migration logic

shintaro-iwasaki:
This PR removes the complicated synchronization code behind `ABT_thread_migrate_to_sched()`, `ABT_thread_migrate_to_xstream()`, and `ABT_thread_migrate()`.

## Background

`ABT_thread_migrate_to_sched()`, `ABT_thread_migrate_to_xstream()`, and `ABT_thread_migrate()` basically find a pool of the running scheduler (associated with the given execution stream) and creates a request of migration to the pool. The request is attached to a ULT since that ULT might be being scheduled by a scheduler. A scheduler who finds a request attached to a ULT will migrate this ULT to the target pool.

## Problems

Obviously, all the involving objects must be alive during the process (i.e., the target pool may not be freed while a thread is being migrated). This is actually hard to guarantee from the runtime perspective. The current Argobots employ several mechanisms to (incompletely) prevent objects from being released; for example, `ABTI_sched.state`, `ABTI_pool.num_migrations`, and `ABTI_xstream.sched_lock` are maintained only for this purpose. Unfortunately, this migration mechanism is rarely used, so this protection mechanism is not thoroughly debugged. It has several "holes", which can be classified as user's error but Argobots seems to deal with. This behavior is not well documented, so this complicated and not-so-common mechanism just become a heavy burden for code maintenance and optimizations.

## Solutions

This PR explicitly describes that it is a user's responsibility to keep them alive during the migration process. Thanks to this PR, we can remove several complicated synchronization mechanism maintained only for this migration functionality.

## Impact

Note that this does not really break the API/ABI compatibility (much); although it was not documented, it is dangerous if one finishes a scheduler while another thread tries to migrate a thread to that scheduler, although previously some cases might be internally avoided by the protection mechanism above. This significantly improves the performance of a stackable scheduler.

If this PR (or commits around this PR) breaks your program (especially regarding thread migration), please let us know.

### DIFF
--- a/src/include/abti.h
+++ b/src/include/abti.h
@@ -219,7 +219,6 @@ struct ABTI_xstream {
     ABTI_xstream_type type;   /* Type */
     ABTD_atomic_int state;    /* State (ABT_xstream_state) */
     ABTI_sched **scheds;      /* Stack of running schedulers */
-    ABTI_spinlock sched_lock; /* Lock for the scheduler management */
     ABTI_sched *p_main_sched; /* Main scheduler, which is the bottom of the
                                * linked list of schedulers */
     ABTI_sched *p_sched_top;  /* The currently running scheduler. This is the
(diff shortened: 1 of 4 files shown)


shintaro-iwasaki:
test:argobots/all

PR #193, opened 2020-06-09
thread: allow ABT_xstream_schedule_unit() on a normal ULT

shintaro-iwasaki:
This patch enables `ABT_xstream_schedule_unit()` on a non-scheduler ULT. This is a large change because this PR essentially provides a way to implement a scheduler without `ABT_sched`.

Specifically, this allows a normal ULT to "directly" schedule ULTs and tasklets by combining `ABT_pool_xxx()` and `ABT_xstream_schedule_unit()` while, for example, waiting for something. From another viewpoint, one can make a scheduler that can also work as a normal ULT (e.g., calling a barrier etc).

This feature should be useful to implement a complicated scheduler, for example, a parallel work unit that does both scheduling and computation. For example, OpenMP thread is a scheduler of OpenMP tasks (belonging to that parallel region) but can perform computation (e.g., `#pragma omp for`).

Internally, this changes the concept of "stackable schedulers" in Argobots. Now it becomes "stackable threads"; any thread can create a "child" thread by `ABT_xstream_schedule_unit()` and return to the "parent" thread by `ABT_thread_yield()` or `ABT_thread_exit()`.  `test/basic/sched_on_thread.c` demonstrates how this feature can be used.

Note that this PR keeps the current API/ABI, so the existing code (including ones that use stackable schedulers) should work.

### DIFF
--- a/src/arch/abtd_thread.c
+++ b/src/arch/abtd_thread.c
@@ -81,21 +81,8 @@ static inline void ABTD_thread_terminate(ABTI_xstream *p_local_xstream,
 
     /* No other ULT is waiting or blocked for this ULT. Since a context does not
      * switch to another context when it finishes, we need to explicitly switch
-     * to the scheduler. */
-    ABTI_sched *p_sched;
-#ifndef ABT_CONFIG_DISABLE_STACKABLE_SCHED
-    if (p_thread->p_sched) {
-        /* If p_thread is a scheduler ULT, we have to context switch to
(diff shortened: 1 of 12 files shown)


shintaro-iwasaki:
test:argobots/all

shintaro-iwasaki:
test:argobots/all

shintaro-iwasaki:
test:argobots/all

PR #196, opened 2020-06-11
key: optimize ABT_key operations

shintaro-iwasaki:
It is related to #159.

This PR optimizes the current `ABT_key` mechanism. Specifically, this PR removes the following:
- Modulo (`%`)
- A reference counter (`refcount`) that is atomically incremented/decremented.

This PR also cleans up `src/key.c`.

PR #197, opened 2020-06-12
key: use memory pool for allocation

### LINKED ISSUE #159: Optimize thread/task TLS mechanism
`ABT_key_set()` and `ABT_key_get()` are costly especially when thread granularity is not coarse:
- When a thread/task accessed TLS for the first time, they call `malloc` twice to allocate `ABTI_ktable` and `ABTI_ktelem` and perform atomic operations on `ref_count` of `ABTI_key`.
- When a thread/task that has used TLS is freed, they call `free` multiple times (# of accessed keys + 1) and perform atomic operations on `ref_count` of `ABTI_key`.
- Argobots does not provide any way to speed it up by, for example, preallocating memory regions.

To avoid this, some programs (for example, BOLT 1.0rc2) relies on `ABT_self_set_arg()` and `ABT_self_get_arg()`. However, this workaround easily arises a compatibility issue (e.g., when two libraries use this feature) and thus is not recommended. `ABT_self_set_arg()` should be removed in the future anyway.

The performance of `ABT_key_set()` and `ABT_key_get()` should be optimized so that programs (for example, BOLT) can use them instead of the current lightweight workaround.

shintaro-iwasaki:
This PR, with #196, fixes #159.

## Problems

Too many `malloc()` in `ABT_key` operations significantly degrade the performance, discouraging users from using the `ABT_key` interface.

## Solutions

`malloc()` is replaced by a generic memory pool introduced by #183. This memory pool is shared with the task descriptor pool, so it does not increase the memory footprint (and does not overly make the current memory pool management).

## Impact

The following shows the overhead [ns] of `ABT_key` operations on 56-core (2-socket Intel Xeon Platinum 8180 Processor). Lower is better. The blue one is before the optimization of #196, the orange bar is the current master (#196), and the gray one is this PR. Basically each ES forks and joins 256 threads (=ULTs) in total, while each ULT accesses the TLS several times. Striped bars mean a large error. It does not show the error bar, but overall the performance on 1ES was stable.

img

This shows that:
- This PR considerably reduces overheads.
- After this PR, the performance on this Skylake (or a typical enterprise Intel server) will be:
  - A new entry creation (i.e., the first access with a certain key) costs 50~100 ns (about 200 cycles).
  - If the key entry already exists in ULT (i.e., subsequent access with the same key), the access cost is around 20 ns (about 50 cycles).

I have not found any issues except for the complicated memory management code. I think this solution is reasonable for now.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -751,6 +751,20 @@ AS_IF([test "x$have_builtin_expect" = "xyes"],
       [AC_DEFINE(ABT_CONFIG_HAVE___BUILTIN_EXPECT, 1,
                  [Define to 1 if you have the `__builtin_expect' function.])])
 
+# check __alignof__
+AC_TRY_COMPILE([], [int a = __alignof__(double); (void)a;],
+               [have_alignof=gcc],
+               [AC_TRY_COMPILE([], [int a = alignof(double); (void)a;],
+                               [have_alignof=c11],
+                               [have_alignof=no])]
+)
+AS_IF([test "x$have_alignof" = "xgcc"],
+      [AC_DEFINE(ABT_CONFIG_HAVE_ALIGNOF_GCC, 1,
+                 [Define to 1 if you have the `__alignof__' operator.])])
+AS_IF([test "x$have_alignof" = "xc11"],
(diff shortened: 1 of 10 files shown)


shintaro-iwasaki:
test:argobots/all

Issue #198, opened 2020-06-12
Replacement of ABT_xxx_set_arg() as TLS


shintaro-iwasaki:
`ABT_self_set_arg()` and `ABT_thread_set_arg()` are used as TLS in some programs since the previous `ABT_key` implementation was poor (#159). This practice is obviously bad since it can easily arise an interoperability issue (e.g., two libraries use this feature); this is also not an intended usage.

The performance issue has been solved by #196 and #197, but `ABT_key` is different from  `ABT_xxx_set_arg()` since `ABT_key_get/set()` can only modify "self" (i.e., the underlying work unit) while `ABT_thread_set_arg()` can target another thread.

Although this use is not recommended at all, we could provide an interface to get/set some data attached to any work unit (i.e., not the currently running unit) by, for example, extending the `ABT_key` interface.

shintaro-iwasaki:
#201 solved this issue.

PR #199, opened 2020-06-13
add a tool interface for profiling

shintaro-iwasaki:
## Backgrounds

Now Argobots does not provide any interface to check the internal state, so it is hard to profile or debug Argobots without modifying the Argobots runtime. As done by MPI and OpenMP (e.g., PMPI and OMPT), Argobots should expose a similar interface. A profiling interface (or, actually a profiling tool) has been demanded by several users (recently from @roblatham00), it would be nice to implement such now.

## Solutions

Argobots provides an interface ("tool interface") to register callback functions, which are called on several threads and tasklets events (e.g., creation, run, join, yield, ... ). The interface is work-unit centric, so with this feature, users can know  avg/max/min execution time of threads (i.e., thread granularity), number of yields and suspensions, delay of scheduling (i.e., time of the first run - time of the creation) and so on with this interface.

## Current issues

The obvious problem is that there exists no such a profiling tool, so I am not sure whether the current set of events is satisfactory or not.

~In any case, working examples should be bundled with Argobots, but the thread-specific storage issue (#198) is the biggest bottleneck for a scalable and reasonably practical profiler; specifically, a profiler needs to save timestamps somewhere per work unit, but getting such a memory location in a scalable manner (`ABT_thread` -> storage: for example, highly scalable TLS with an extended API or highly scalable parallel hash-table / parallel search tree) is not easy with the current Argobots implementation.~ [EDIT] This specific issue was solved.

This PR should be merged after confirming a few profilers work well.

### Note

This feature definitely adds an overhead, so this is disabled by default (`--enable-tool` is needed). CI has been already set up, so we will test this feature regularly. The current code is messy, so it should require refactoring in the future. 

Any suggestions or comments are welcome.

### [EDIT] Performance Impact

The following shows the overhead of ULT fork-join, which is one of the representative performance-critical operations in Argobots. The benchmark repeats fork-join N ULTs (N = X axis) on multiple execution streams. The error bar was not measured.

The blue bars show the performance of the current master. The orange bars show the performance of this PR (without the tool feature, which is the default behavior). The gray shows the overheads when the tool feature is enabled (which must be specified by a user). I wanted to show the following
1. Blue = Orange: with the default configurations, this PR does not change the performance significantly.
2. Gray > Orange and Blue: the tool feature can slightly increase the overheads. 

In reality, however, I can hardly say that there is a performance difference from the results without error bars. In any case, the performance difference is similar and acceptable.
#### Intel Xeon Platinum 8180
Intel Xeon Platinum 8180
#### Intel Xeon Phi 7210 (KNL)
Intel Xeon Phi 7210 (KNL)

#### Summit-like POWER 9
Summit-like POWER 9

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -196,6 +196,11 @@ AC_ARG_ENABLE([dynamic-promotion],
                    [use the dynamic promotion threading technique that performs
                     better if a ULT does not yield.]))
 
+# --enable-tool
+AC_ARG_ENABLE([tool],
+    AS_HELP_STRING([--enable-tool],
+                   [enable the tool interface, which is disabled by default.]))
+
 # --disable-simple-mutex
 AC_ARG_ENABLE([simple-mutex],
     AS_HELP_STRING([--disable-simple-mutex], [use an experimental mutex implementation]),,
@@ -620,6 +625,10 @@ AS_IF([test "x$enable_dynamic_promotion" = "xyes" -a "x$enable_fcontext" != "xno
                  [Define to use the dynamic promotion technique for ULT])],
       [AC_DEFINE(ABT_CONFIG_THREAD_TYPE, ABT_THREAD_TYPE_FULLY_FLEDGED)])
 
+# --enable-tool
+AS_IF([test "x$enable_tool" != "xyes"],
+      [AC_DEFINE(ABT_CONFIG_DISABLE_TOOL_INTERFACE, 1,
+                 [Define to use the tool interface])])
 
 # check if __atomic builtins are supported
 AC_TRY_COMPILE(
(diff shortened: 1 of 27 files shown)


shintaro-iwasaki:
test:argobots/all

shintaro-iwasaki:
test:argobots/all

shintaro-iwasaki:
test:argobots/all

shintaro-iwasaki:
test:argobots/all

shintaro-iwasaki:
test:argobots/all

shintaro-iwasaki:
This feature passes all the tests including newly added test configurations for tools.
This feature is still experimental, but I believe this PR is worth adding.

PR #201, opened 2020-06-16
key: extend get/set_specific interface

shintaro-iwasaki:
## Problems

This addresses #198; it would be nice to have TLS functions that can be accessed by not only the owner (=self) but also other parallel units as well.

## Solutions

This PR implements the following functions:
```c
int ABT_thread_set_specific(ABT_thread thread, ABT_key key, void *value);
int ABT_thread_get_specific(ABT_thread thread, ABT_key key, void **value);
int ABT_task_set_specific(ABT_task task, ABT_key key, void *value);
int ABT_task_get_specific(ABT_task task, ABT_key key, void **value);
```
These functions take a work unit handle, so these functions can access other's TLS.

## Performance impact

It degrades the performance of `ABT_key` access since atomic operations are newly added in some paths. In Argobots, a key table and its element of a key table are lazily created, so synchronization is necessary if two parallel units try to allocate a key table or insert a new element (i.e., two parallel `ABT_xxx_set_specific()`).

The following is the overheads (nanoseconds) of `ABT_key` operations on some CPUs: 56-core Skylake (2-socket Intel Xeon Platinum 8180 Processor), 64-core Intel Knights Landing 7210, and Summit-like POWER 9. I also measured the performance on a nice ARM CPU. Lower is better.

The blue one is before the optimization of #196, the orange bar is #196, the gray one is #197, and the yellow one shows this PR. Basically each ES forks and joins many ULTs (256 for 1ES and 4096 for more ESs) in total, while each ULT accesses **its own TLS** several times (so no contention). It shows the average of five time measurement.

img2

The result shows that, yes, this new synchronization adds certain overheads (10% - 80%)), but I believe this cost is acceptable enough considering the original performance of `ABT_key` operations. Specifically, this feature is useful for #199 but also contributes to #198. The performance on 64-bit ARM was almost similar to the other three CPUs.

Note that basically the performance is degraded only on the first entry creation; subsequent TLS access with the same `ABT_key` does not get costlier.

### DIFF
--- a/src/include/abt.h.in
+++ b/src/include/abt.h.in
@@ -577,6 +577,8 @@ int ABT_thread_get_stacksize(ABT_thread thread, size_t *stacksize) ABT_API_PUBLI
 int ABT_thread_get_id(ABT_thread thread, ABT_unit_id *thread_id) ABT_API_PUBLIC;
 int ABT_thread_set_arg(ABT_thread thread, void *arg) ABT_API_PUBLIC;
 int ABT_thread_get_arg(ABT_thread thread, void **arg) ABT_API_PUBLIC;
+int ABT_thread_set_specific(ABT_thread thread, ABT_key key, void *value) ABT_API_PUBLIC;
+int ABT_thread_get_specific(ABT_thread thread, ABT_key key, void **value) ABT_API_PUBLIC;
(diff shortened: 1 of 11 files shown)


shintaro-iwasaki:
test:argobots/all

shintaro-iwasaki:
test:argobots/all

PR #202, opened 2020-06-16
Remove inappropriate terms in the code

shintaro-iwasaki:
This PR updates some terms in the Argobots code.

We also change the default branch of Argobots to `main`; we will keep `master` updated for a week or so but plan to remove the branch after that.


shintaro-iwasaki:
test:argobots/all

PR #203, opened 2020-06-18
thread/task: add functions to check if a work unit is unnamed

shintaro-iwasaki:
This patch adds utility functions that check if a given work unit is unnamed (or anonymous) or not.

```c
int ABT_thread_is_unnamed(ABT_thread thread, ABT_bool *flag);
int ABT_task_is_unnamed(ABT_task task, ABT_bool *flag);
int ABT_self_is_unnamed(ABT_bool *flag);
```

They should be useful in general, but they are needed specifically for #199 (precisely speaking, for tool tests) because a state transition pattern of a named work unit is different from that of an unnamed work unit (e.g., no "join").


shintaro-iwasaki:
test:argobots/all

shintaro-iwasaki:
test:argobots/all

PR #205, opened 2020-07-08
pool: add ABT_pool_pop_wait()

shintaro-iwasaki:
## Problems

`ABT_pool_pop_timedwait()` is a pool function that can sleep an underlying execution stream when there is no work unit in the pool. This function takes an absolute time as do `pthread_cond_timedwait()` and other condition variables (since it uses a condition variable inside), but getting absolute time is not very lightweight although this wait does not always happen (unlike `pthread_cond_timedwait()`). Consider the following case:
```c
void sched_run() {
   while (1) {
     ABT_unit unit;
     ABT_pool_pop_timedwait(pool, &unit, ABT_get_wtime() + 0.1); /* wait 0.1 secs if the pool is empty */
     if (unit != ABT_UNIT_NULL) [...];
   }
}
```
This `ABT_get_wtime() + 0.1` is not used if the pool has a work unit, but the user always needs to pass this value.  `ABT_get_wtime()` is heavy (this is mapped to a system call), so it can add considerable overheads.

This interface also forces (user-defined) pool implementation to use this system absolute time (e.g., `ABT_get_wtime()`) although the pool implementation might prefer another timer (e.g., `gettimeofday()` or `rdtscp`).

## Solution

This PR implements `ABT_pool_pop_wait()`, which takes a relative wait time. The code will be like this:
```c
void sched_run() {
   while (1) {
     ABT_unit unit;
     ABT_pool_pop_wait(pool, &unit, 0.1); /* wait 0.1 secs if the pool is empty */
     if (unit != ABT_UNIT_NULL) [...];
   }
}
```
The user does not need to call `ABT_get_wtime()` every time before calling `ABT_pool_pop_wait()`. The pool implementation may omit `ABT_get_wtime()` if the pool is not empty. This patch marks the old timedwait interface as deprecated.

## Expected impact

This PR basically implements a new feature, so it does not affect the performance. This PR should slightly improve the performance of `BASIC_WAIT` scheduler, though. Even after this PR, Argobots will keep providing `ABT_pool_pop_timedwait()` for a while.

This patch, however, breaks the compatibility since it adds a new parameter to `struct ABT_pool_def`. It would be a compile-time error if a user statically initializes `ABT_pool_def`-typed variable in the program.
```c
ABT_pool_def pool_def = {x_func, y_func, ..., }; /* compile error since it has a new parameter*/
```
I do not think this becomes a major issue. This PR will be merged into Argobots 1.0 series.


shintaro-iwasaki:
test:argobots/all

shintaro-iwasaki:
test:argobots/osx

shintaro-iwasaki:
test:argobots/freebsd

shintaro-iwasaki:
test:argobots/all
test:argobots/osx
test:argobots/freebsd

src/pool/fifo_wait.c
```diff
@@ -132,6 +134,65 @@ static void pool_push(ABT_pool pool, ABT_unit unit)
...
+{
+    ABTI_pool *p_pool = ABTI_pool_get_ptr(pool);
+    data_t *p_data = pool_get_data_ptr(p_pool->data);
+    unit_t *p_unit = NULL;
+    ABT_unit h_unit = ABT_UNIT_NULL;
+
+    pthread_mutex_lock(&p_data->mutex);
+
+    if (!p_data->num_units) {
+#if defined(ABT_CONFIG_USE_CLOCK_GETTIME)
```
carns, line 147: Just for my own education: on what platforms might we expect ABT_CONFIG_USE_CLOCK_GETTIME to be undefined?  I like that there is a fallback here, but it would be good to know when it might happen because there will be a performance implication.

shintaro-iwasaki: I believe `ABT_CONFIG_USE_CLOCK_GETTIME` is enabled on 100% (or almost 100%) platforms. This branch is to make sure Argobots can be compiled in a special case.

This check comes from https://github.com/pmodels/argobots/blob/main/src/arch/abtd_time.c , which I believe should stem from MPICH's code (`MPI_Wtime()`), but I could not find platforms that do not support `clock_gettime()` (except for Windows, which Argobots does not support).

If such a platform would exist, there should be a way to get an absolute time for `pthread_cond_timedwait()`, so once we find such, we will fix this implementation.

carns: Ah Ok.  Glad to hear that it's not expected to be a common scenario.   Thanks!

PR #215, opened 2020-07-27
thread: fix a bug in creation of multiple unnamed ULTs

shintaro-iwasaki:
This PR fixes a bug in `ABT_thread_create_many()`.

`ABT_thread_create_many()` passes `NULL` to `ABTI_thread_create_internal()`'s `pp_newthread` when it needs to create an unnamed ULT while `pp_newthread` may not be `NULL`. This PR fixes it.


shintaro-iwasaki:
test:argobots/all

PR #216, opened 2020-07-28
test/basic: fix memory leaks

shintaro-iwasaki:
This PR fixes memory leaks in `test/basic`. Note that this bugs are not in the Argobots library but in the test programs.


shintaro-iwasaki:
test:argobots/all

PR #221, opened 2020-08-12
Remove ABTI_unit

shintaro-iwasaki:
## Background

The current structures of `ABTI_unit`, `ABTI_thread`, and `ABTI_task` are as follows:
```c
struct ABTI_unit {
    [...]; // Common variables
};
struct ABTI_thread { // ULT descriptor
    ABTI_unit unit_def;
    [...]; // ULT-specific variables
};
struct ABTI_task { // tasklet descriptor
    ABTI_unit unit_def;
};
```
Obviously, there is no tasklet-specific data, so `ABTI_task` is equal to `ABTI_unit`. This composition makes the code lengthy and dirty.

## This PR
This PR changes the structures as follows.
```c
struct ABTI_thread { // descriptor for both ULT and tasklet
    [...]; // Common variables
    ABTI_thread_context ctx;
};
struct ABTI_thread_context {
    [...]; // ULT-specific variables
};
```
`ABTI_thread` is used for both ULTs and tasklets, but `ABTI_thread` for tasklets is smaller; it does not allocate the `ABTI_thread_context` part in `ABTI_thread` so that a tasklet descriptor fit in two 64-byte cache lines. Consequently, this PR renames/removes `unit`-related variables and functions.

### Expected Impact

This PR only changes the internal implementation (for example, this PR keeps `ABT_unit`) while we will integrate tasklets and ULTs more. The performance change should be negligible.

### DIFF
--- a/src/arch/abtd_thread.c
+++ b/src/arch/abtd_thread.c
@@ -12,13 +12,12 @@ void ABTD_thread_func_wrapper(void *p_arg)
 {
     ABTD_thread_context *p_ctx = (ABTD_thread_context *)p_arg;
     ABTI_thread *p_thread = ABTI_thread_context_get_thread(p_ctx);
-    ABTI_xstream *p_local_xstream = p_thread->unit_def.p_last_xstream;
+    ABTI_xstream *p_local_xstream = p_thread->p_last_xstream;
     ABTI_tool_event_thread_run(p_local_xstream, p_thread,
-                               p_local_xstream->p_unit,
-                               p_thread->unit_def.p_parent);
-    p_local_xstream->p_unit = &p_thread->unit_def;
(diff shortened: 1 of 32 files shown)


shintaro-iwasaki:
test:argobots/all

shintaro-iwasaki:
test:argobots/all

shintaro-iwasaki:
On second thought, I believe we should distinguish ULT (or unyieldable thread) from tasklet (unyieldable thread) by type for better code maintenance. Actually it should be done by inheritance, but C90 does not have such a feature (unnamed struct could be one option but it is from C11), so it should be addressed by composition ("a child struct owns a parent struct"). I will create another PR.

shintaro-iwasaki:
Supersede by #228

PR #224, opened 2020-08-18
Always enable the stackable scheduler feature

shintaro-iwasaki:
This PR refactors code for a stackable scheduler. As a result, a stackable scheduler can be always enabled without performance loss.

Note that a stackable scheduler was enabled by default, so this change does not affect most users (except for small overhead reduction).


shintaro-iwasaki:
test:argobots/all

PR #225, opened 2020-08-18
unit: embed refcount and migratable

shintaro-iwasaki:
This PR embeds `refcount` and `migratable` in `ABTI_unit_type`. This can make the struct size of `ABTI_unit` smaller. This change is necessary for the future commit.

This PR does affect the existing code.


shintaro-iwasaki:
test:argobots/all

PR #227, opened 2020-08-20
task: unify task and thread

shintaro-iwasaki:
## Background

`ABT_task` and `ABT_thread` are two different work units in Argobots, but their functionalities are very similar. In practice, `ABT_thread` is a superset of `ABT_task`; for example, `ABTI_task` is `ABTI_unit` while `ABTI_thread` has `ABTI_unit` and other variables to preserve ULT context. Unfortunately, the current Argobots abstraction does not capture this fact, so all the functions that deal with "work units" have two versions for task and thread. This causes API-level inconsistency and makes code maintenance complicated.

## Solution

This PR tries to internally unify task and thread and removes the redundant implementation as much as possible. `ABT_thread` can be either a ULT or a tasklet. A ULT-type work unit is internally treated as a yieldable thread ("ythread"). This can also contribute to #170 since the semantics gets clearer; both "ULT" and "tasklet" are of `ABT_thread` type while "tasklet" is a thread that cannot yield (which seems very obvious but it was not from the API perspective).

## Expected impact

This PR breaks the ABI compatibility but keeps the API compatibility.
This PR increases the overheads since adds branches to handle tasklet in thread functions. The performance impact is not very significant; I would like to prioritize code maintenance.

Note that this change is quite large (it changes 3000 lines of code while the Argobots library has 30000 lines).

The following is the fork-join overheads of tasklets and yieldable threads. The benchmark forks N threads (X-axis) and joins all of them. The benchmark uses one execution stream.
image


shintaro-iwasaki:
test:argobots/all

shintaro-iwasaki:
test:argobots/all

shintaro-iwasaki:
This change is quite large. If this change breaks your program, please let us know.

PR #228, opened 2020-08-20
rename ABTI_thread, ABTI_task, and ABTI_unit

shintaro-iwasaki:
This PR renames the structs of work units. Specifically, this PR changes `ABTI_thread`, `ABTI_task`, and `ABTI_unit` as follows:
```c
struct ABTI_unit {
    [...]; // Common variables
};
struct ABTI_thread { // ULT descriptor
    ABTI_unit unit_def;
    [...]; // ULT-specific variables
};
struct ABTI_task { // tasklet descriptor
    ABTI_unit unit_def;
};
```

This PR changes these structures as follows.
```c
struct ABTI_thread { // descriptor for both ULT and tasklet
    [...]; // Common variables
};
struct ABTI_ythread { // ULT-specific descriptor
    ABTI_thread thread;
    [...]; // ULT-specific variables
};
```
The point of this change is to make it clear that both ULTs and tasklet are "threads" (`ABTI_thread`) while ULTs are special threads in terms of yieldability (thus yieldable thread: `ABTI_ythread`). This relation should be elegantly represented with inheritance, but this PR uses composition instead since C does not support such.

This is prerequisite for #227.

### DIFF
--- a/maint/template.c
+++ b/maint/template.c
@@ -5,7 +5,7 @@
 
 #include "abti.h"
 
-int ABTI_template(ABTI_thread *p_thread, void *p_arg)
+int ABTI_template(ABTI_ythread *p_thread, void *p_arg)
 {
     int abt_errno = ABT_SUCCESS;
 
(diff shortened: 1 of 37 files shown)


shintaro-iwasaki:
test:argobots/all

PR #231, opened 2020-08-21
refactoring: external thread check

shintaro-iwasaki:
This PR cleans up the code regarding the external thread feature.


shintaro-iwasaki:
test:argobots/all

PR #234, opened 2020-08-25
util: remove ABTU_get_int_len()

shintaro-iwasaki:
Only `ABTU_get_int_len()` uses on a math library in Argobots, but this function is not very necessary.  This patch removes it.

Note that it can remove the math library dependency (`-lm`) from Argobots.


shintaro-iwasaki:
test:argobots/all
test:argobots/osx
test:argobots/freebsd
test:argobots/solaris

PR #235, opened 2020-08-26
thread: improve ABTI_thread_create() and ABTI_thread_free()

shintaro-iwasaki:
This patch improves functions called in `ABTI_thread_create()` and `ABTI_thread_free()`.

Note that `sizeof(ABTI_ythread)` on most architectures including x86/64, POWER9, and 64-bit ARM becomes 128 bytes after this patch (*if `fcontext` is enabled, a compiler is GCC compatible in terms of structure alignment, and the pointer size is 8 bytes (64-bit OS)).

### DIFF
--- a/src/include/abtd_ythread.h
+++ b/src/include/abtd_ythread.h
@@ -17,28 +17,22 @@ void ABTD_ythread_func_wrapper(void *p_arg);
 void ABTD_ythread_terminate_no_arg();
 #endif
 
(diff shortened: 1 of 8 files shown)


shintaro-iwasaki:
test:argobots/all

shintaro-iwasaki:
test:argobots/all

shintaro-iwasaki:
test:argobots/all

Issue #116, opened 2019-09-19
malloc failures should be checked and assertion should be avoided


jolivier23:
Malloc can fail normally when there is memory pressure.   There may be situations where it is appropriate to exit the application but in many cases garbage collection or other techniques can be used to free memory and recover appropriately.  In our project (DAOS), we generally want to return an error code to the client in such situations and we want the opportunity to take steps in the server (where we use argobots) to reduce memory consumption and avoid future malloc failures at runtime rather than killing the server as the latter is a very expensive operation in a fault tolerant storage system.

This ticket is a feature request to return error codes on out-of-memory situations, at least at the API level so that users of the argobots library can choose whether to attempt to free up memory and try to recover or to give up and crash.  This is related to ticket #115 but not specific to just the valgrind code path so adding as a new ticket

shintaro-iwasaki:
Resource allocation failures in the current Argobots is reasonably, though not thoroughly (e.g., by using formal verification techniques), tested. Argobots functions return an error properly after freeing resources already allocated. The new tests introduced by #309 check error handling paths by mocking system allocation functions (`malloc()`, `mmap()`, `pthread_create()`, and so on).

Please open an issue if you find any issue.

PR #238, opened 2020-08-27
mem: return an out-of-memory error to users

shintaro-iwasaki:
This PR (partially) addresses #116.

Argobots often ignores a return value of `malloc()`, which can cause SEGV when a system does not have enough memory. This PR partially fixes it by forcing a caller to check an error code so that a caller can return an error to the user.

Note that this PR does not fully address the issue mentioned by #116. In particular, some large initialization functions such as `ABT_init()` and `ABT_xstream_create()` assume that resource allocation always succeeds, although most smaller functions (e.g., `ABT_thread_create()` now properly handle `malloc()` and its friends. Fixing such large functions is future work (and I believe they are less likely to fail compared with functions that are called repeatedly after initialization of larger modules/objects).

Argobots tests do not fully check the memory allocation error; to check this, we might need to create unit tests.

This PR potentially adds branches if function inlining does not work well, but this change is necessary for safety and correctness.


shintaro-iwasaki:
test:argobots/all
test:argobots/osx

PR #239, opened 2020-08-28
pool: remove producer consumer check

shintaro-iwasaki:
### Issue 1: producer/consumer error checks do not work
The current pool producer/consumer design is not very effective.  Specifically, Argobots does not properly handle producer/consumers on `ABTI_ythread_yield()` (the parent thread needs to push a thread that called `ABTI_ythread_yield()` to its associated pool) and `ABTI_ythread_set_ready()` (`ABTI_ythread_set_ready()` pushes a suspended thread to its associated pool), both of which are heavily used in Argobots. In reality, Argobots misses several error checks regarding pool producer/consumer.

### Issue 2: only execution streams can be a producer/consumer
In the current design producers and consumers must be execution streams, which is very restrictive. For example, a pool with a limited access does not work with external threads or migratable schedulers (e.g., scheduler threads).  Potentially, however, such a pool can work with single-producer and/or single-consumer pools under certain (user-customized) scheduling (for example, in a case where only one "thread" (not an execution stream) accesses a pool).

## What this PR does

This PR remove this incomplete and restrictive produce/consumer checks from Argobots.

### Discussion: possible solutions

One (I mean, I) can come up with several solutions/workarounds:
1. Ignore all access-related errors and use an access mode as an implementation hint

2. Always check a producer/consumer error

    2.1. Check an error on actual `yield`/`set_ready`: it means that synchronization operations (including `mutex_unlock()`/`cond_broadcast()`/`barrier_wait()`) can fail because of pool access mode.
    2.2. Check an error on yield/suspend: it means that synchronization operations (including thread_join/mutex_lock/cond_wait) will 
use yield-based synchronization conservatively if the underlying pool only allows a single producer.
   - In both cases, users cannot handle such an error (= an error is fatal).

3. Allow multiple producers even for single-producer pools by implementing a mechanism with which other producers can ask a main producer to push a thread.
3.1. If a producer is an execution stream, this request handling is done in the main scheduler. This is possible but restrictive.
3.2. If a producer can be one of a thread, an external thread, and an execution stream, deciding where to handle a request is challenging.
    - To flexibly handle producer/consumer information, users might need to change it dynamically, which I think is impractical since it is too burdensome. In addition, it kills the merit of using single-producer/consumer pools since this mechanism will work as a multiple-producer/consumer queuing system.

I believe 1 is preferred because both 2 and 3 are incomplete and/or restrictive in any case while they add certain overheads.  As a low-level threading library, user needs to be responsible for it(=no error check at rutnime).  The pool access modes is considered to be a hint attribute (e.g., `MPI_THREAD_SERIALIZED` in MPI).

At present only the migration feature and `ABT_thread_join()` change their algorithm based on this access mode.  We keep their algorithms for compatibility. In the future, we can create yield-based versions of synchronization objects and otherwise use an optimal algorithm that assumes MPMC pools in the future.

### Expected Impact

Performance will be improved since it removes an error check. It can cause an error if programs rely on this (incomplete) producer/consumer check, but this is less likely; please let us know if this change breaks your program.

Note that there is no problem if MPMC pools are used.


shintaro-iwasaki:
test:argobots/all

shintaro-iwasaki:
test:argobots/all

shintaro-iwasaki:
test:argobots/all

PR #241, opened 2020-08-30
barrier: fix a memory bug in ABT_barrier_reinit()

shintaro-iwasaki:
This PR fixes a bug introduced by #238.


shintaro-iwasaki:
test:argobots

PR #242, opened 2020-08-31
xstream: introduce a root thread

shintaro-iwasaki:
## Problem

Main schedulers are running on yieldable `ABTI_thread`, but they cannot yield because these schedulers do not have parent threads. Main schedulers, therefore, cannot run synchronization operations that might call yield; for example, main schedulers cannot call `ABT_mutex_lock()`, `ABT_thread_yield()` and `ABT_thread_join()`.

This restriction was not clearly explained.  Some functions return an error code while others silently cause a fatal error.  For proper error handling, this behavior must be clarified.

## What this PR does.

This PR introduces a root thread that schedules a main scheduler.  This fix makes all the user-visible "ULT"-based threads (or yieldable-threads, or user-visible `ABTI_ythread`) truly yieldable.  This fix cleans up some existing code as well.

Nonetheless, users are not recommended to block main schedulers at all; it can easily cause a deadlock. Please run these operations on main schedulers only if you fully understand what scheduling would happen when some main schedulers block.

Note that root threads cannot yield since they do have parents, but users are not aware of it since they are not visible to users (maybe using a tool interface it can be exposed, which should be fixed in the future),

### The other possible solution

I did not choose the other possible solution, which prevents main schedulers from executing all the synchronization functions by adding a flag check because it makes user-visible yieldable threads virtually non-yieldable. For clearer semantics, it is better to make all the yieldable threads yieldable.

### Performance/API Impact

The performance/API impact is negligible.  The error handling has been modified, but I do not think any users rely on error codes returned from blocking operations on main schedulers that are customized by users.  This PR, however, significantly changes the internal logic of execution stream state management, so it might introduce a new bug (although this PR has been carefully tested).  Please let us know if this PR breaks your code.

### DIFF
--- a/src/arch/abtd_ythread.c
+++ b/src/arch/abtd_ythread.c
@@ -42,7 +42,8 @@ static inline void ythread_terminate(ABTI_xstream *p_local_xstream,
         /* If p_link is set, it means that other ULT has called the join. */
         ABTI_ythread *p_joiner = ABTI_ythread_context_get_ythread(p_link);
         if (p_ythread->thread.p_last_xstream ==
-            p_joiner->thread.p_last_xstream) {
+                p_joiner->thread.p_last_xstream &&
+            !(p_ythread->thread.type & ABTI_THREAD_TYPE_MAIN_SCHED)) {
             /* Only when the current ULT is on the same ES as p_joiner's,
              * we can jump to the joiner ULT. */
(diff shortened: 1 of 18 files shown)


shintaro-iwasaki:
test:argobots/all

shintaro-iwasaki:
test:argobots/all

PR #243, opened 2020-08-31
error: improve error handling

### LINKED ISSUE #96: Memory leak on error
Memory is leaked when error handling is done internally.

For example, `ABT_sched_create` (in https://github.com/pmodels/argobots/blob/master/src/sched/sched.c) does not free `p_sched`, `pool_list`, and possibly some `ABTI_pool`s if pool creation fails.
Such a minor memory leak also happens in `ABTI_cond_wait`. Although these unusual paths are not taken in a sound program, we should examine all functions.

Static analysis (e.g., LLVM's static analyzer) is helpful.

shintaro-iwasaki:
This PR addresses #116. This PR improves error handling (including out-of-memory errors) of functions that are frequently used in user programs. Specifically, all the functions that might return errors will be decorated with `warn_unused_return`, so Argobots developers are more aware of error handling.  This PR, however, does not fix all the functions.

1. When `ABT_init()`, `ABT_finalize()`, and `ABT_xstream_create()` fail, the runtime state becomes unstable.
2. Some functions called in `ABT_init()` do not handle out-of-memory errors.
3. `ABT_thread_create_many()`, `ABT_thread_join_many()`, and `ABT_thread_free_many()` are not atomic; they can partially create, join, or free threads.

I believe, however, this patch covers most functions.  Because the remaining functions are large and complicated, further improvement requires tests that can check error paths, which is the future work but so far of lower priority.

This PR fixes #96.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -273,6 +273,8 @@ fi
 AC_SUBST([ABT_DEPRECATED])
 # check __attribute__((noreturn))
 AX_GCC_FUNC_ATTRIBUTE(noreturn)
+# check __attribute__((warn_unused_result))
+AX_GCC_FUNC_ATTRIBUTE(warn_unused_result)
 
 dnl ----------------------------------------------------------------------------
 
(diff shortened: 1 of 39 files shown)


shintaro-iwasaki:
test:argobots/all
test:argobots/osx
test:argobots/freebsd
test:argobots/solaris

shintaro-iwasaki:
test:argobots/all

Issue #244, opened 2020-08-31
Another XLC compiler bug


shintaro-iwasaki:
Another XLC bug (former one: #162).

IBM XL C/C++ 16.01 compiler cannot inline functions properly in some cases.
```c
#include <stdlib.h>
static inline int malloc1(void **p_ptr) {
    void *ptr = malloc(sizeof(int));
    if (ptr == 0)
        return 1;
    *p_ptr = ptr;
    return 0;
}
static inline int malloc2(void **p_ptr) {
    return malloc1(p_ptr);
}
int main() {
    int *ptr;
    malloc2((void **)&ptr);
    *ptr = 1; // <--- this ptr is not updated
    return 0;
}
```
In this case, `ptr` is not updated in `main()`, which causes SEGV.
```
$ xlc --version
IBM XL C/C++ for Linux, V16.1.1 (Community Edition)
Version: 16.01.0001.0003
$ xlc -O3 code.c
$ valgrind ./a.out
==137751== Memcheck, a memory error detector
==137751== Copyright (C) 2002-2017, and GNU GPL'd, by Julian Seward et al.
==137751== Using Valgrind-3.14.0 and LibVEX; rerun with -h for copyright info
==137751== Command: ./a.out
==137751== 
==137751== Use of uninitialised value of size 8
==137751==    at 0x10000754: main
==137751== 
```
This compiler bug (note any other compilers+architectures cause this issue) causes an error after da4c1237831bce3f4b2030e987893e7f4d01e72a . It seems that XLC will move to Clang frontend (https://developer.ibm.com/components/ibm-power/blogs/c-and-fortran-adopt-llvm-open-source/), so in the future we can drop this legacy XLC, but we need to fix this for now.

shintaro-iwasaki:
Small correction. This XLC is based on Clang since it defines `__ibmxl__` but no `__xlc__` (https://www.ibm.com/support/knowledgecenter/SSGH3R_16.1.0/com.ibm.xlcpp161.aix.doc/compiler_ref/xlmacros.html, https://sourceforge.net/p/predef/wiki/Compilers/). This implies that even Clang-based one is perhaps not very mature.

Note that this inlining does not cause an issue with genuine Clang (including Clang 3.9.1, Clang7.0.0, and Clang10.0.0 on this POWER 9 machine).

shintaro-iwasaki:
#245 fixes this issue.

PR #245, opened 2020-09-01
work around XLC compilation bug

shintaro-iwasaki:
This patch works around the issue reported in https://github.com/pmodels/argobots/issues/244 by adding a special path for IBM XLC. We can remove it once XLC addresses this issue. I have already confirmed that this solves the issues on POWER9 with XLC 16.1.1.

Note that this does not affect the performance if the other compilers are used.

### DIFF
--- a/src/include/abtu.h
+++ b/src/include/abtu.h
@@ -107,6 +107,19 @@
 
 /* Utility Functions */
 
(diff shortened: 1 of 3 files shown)


shintaro-iwasaki:
test:argobots

PR #247, opened 2020-09-15
error: do not update a variable on failure

shintaro-iwasaki:
## Background

Some Argobots functions return errors for several reasons.  On failure, most Argobots function do not modify values including the Argobots' internal values and arguments given by the user.  This atomic behavior is reasonable: every function should either complete an operation or return an error without any side effect.

A few functions, however, update user-given arguments, which violates this policy and makes the behavior of Argobots functions inconsistent.  For example, `ABT_thread_create()` sets `*thread = ABT_THREAD_NULL` if it is called on any type of failure (e.g., memory allocation failure or calling `ABT_thread_create()` before `ABT_init()`) while `ABT_mutex_create()` does not (`*mutex` will not be modified).  Obviously, Argobots should not modify values on failure; otherwise, it should set `NULL` as `ABT_SUCCESS`.

## Solution

This PR removes side effects on failure so that all functions become atomic.  Unfortunately, it's not complete (e.g., `ABT_init()` might leak memory on failure), but this PR advances it.

Note that there are four functions that return an error code AND change user-given arguments in order to maintain the backward compatibility: `ABT_thread_self()`, `ABT_task_self()`, `ABT_xstream_self()`, `and ABT_self_get_type()`.

## Expected Impact

There is no visible performance impact, but some functions are rewritten not to modify values on failure (although this behavior is not mentioned in the document).  This change can cause a problem if the program relies on such behavior.  Please check a returned error code instead in order to check if there is an error.

If this change breaks the existing code significantly, please let us know so that we can fix it / revert this change.

### DIFF
--- a/src/pool/pool.c
+++ b/src/pool/pool.c
@@ -42,7 +42,6 @@ fn_exit:
     return abt_errno;
 
 fn_fail:
-    *newpool = ABT_POOL_NULL;
     HANDLE_ERROR_FUNC_WITH_CODE(abt_errno);
     goto fn_exit;
 }
@@ -75,7 +74,6 @@ fn_exit:
 
 fn_fail:
     HANDLE_ERROR_FUNC_WITH_CODE(abt_errno);
-    *newpool = ABT_POOL_NULL;
     goto fn_exit;
 }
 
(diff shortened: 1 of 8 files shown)


shintaro-iwasaki:
test:argobots/all

shintaro-iwasaki:
test:argobots/all

shintaro-iwasaki:
test:argobots/all

PR #249, opened 2020-09-24
sched: initialize p_sched->p_data

shintaro-iwasaki:
`ABT_sched`'s `p_data` is not initialized in `ABT_sched_create()`, so `ABT_sched_get_data()` returns an uninitialized value.  This behavior is not intuitive for most users.  Since scheduler creation is not performance critical, this PR initializes `p_data` to NULL on creation.


shintaro-iwasaki:
test:argobots/all

PR #252, opened 2020-10-02
configure: no inlining for IBM XL C/C++ compilers

### LINKED ISSUE #251: XLC compiler bug reappears
This is related to #244. With IBM XL C/C++ compiler, Argobots gets broken because of the #244 issue (specifically, with `--disable-mem-pool --enable-checks=no-error`). It seems that XLC cannot inline functions correctly (which is enabled by "-O2").

Since this bug is extremely annoying and hard to debug, we should completely disable inlining if XLC is used. If we can confirm this bug is fixed in the future, we can update the condition.

shintaro-iwasaki:
This fixes #251 by disabling inlining when `xlc` / IBM XLC compiler is used. There's no performance impact if you use other compilers.


shintaro-iwasaki:
test:argobots/all

shintaro-iwasaki:
test:argobots/all

shintaro-iwasaki:
This passed all the tests on POWER 9 + XLC 16.1.1.

PR #259, opened 2020-10-15
barrier: fix ABT_barrier_free

shintaro-iwasaki:
`ABT_barrier` might be freed right after `ABT_barrier_wait()`.  `lock` is taken in `ABT_barrier_free()` in order not to free `ABT_barrier()` that is being used.  The number of waiters, however, is checked **before** this lock, which causes an error if `assert()` is enabled; before taking `lock`, there might exist some waiters.

This PR fixes it by checking the number of waiters **after** taking `lock`.


shintaro-iwasaki:
test:argobots/all
test:argobots/osx

PR #260, opened 2020-10-22
cond: declare static functions at the bottom of the code file

shintaro-iwasaki:
Internal static functions should be declared after public `ABT_` functions.  This PR fixes it.

This fix is minor; it won't affect the performance of Argobots.


shintaro-iwasaki:
test:argobots/all

shintaro-iwasaki:
test:argobots/all

PR #263, opened 2020-11-12
env: improve reading process of Argobots' environment variables

shintaro-iwasaki:
## Problem

Argobots reads environment variables without checking if values are in an acceptable range.  
Although passing such a large value (e.g., 4GB for page size) is not practical,, some variable type are improper to read and store these values: `int` and `uint32_t` are used for data size. Values are read by 'atoi()' and `atol()`, so they might fail to read big unsigned values.

## What this PR does
This PR cleans up the code in `ABTD_env_init()`; it fixes type issues and uses new `atoi()` functions that can handle overflow.

### DIFF
--- a/src/arch/abtd_env.c
+++ b/src/arch/abtd_env.c
@@ -13,7 +13,6 @@
 #define ABTD_SCHED_EVENT_FREQ 50
 #define ABTD_SCHED_SLEEP_NSEC 100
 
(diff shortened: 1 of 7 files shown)


shintaro-iwasaki:
test:argobots/all

shintaro-iwasaki:
test:argobots/all

shintaro-iwasaki:
test:argobots/all

PR #268, opened 2020-12-02
simplify algorithms of synchronization objects

shintaro-iwasaki:
## Problem

The implementations of synchronization objects are overlapped but differently written in each file. Specifically, the core algorithm of synchronization objects is managing a queue of waiting threads, but `ABT_cond`, `ABT_barrier`, `ABT_future`, and `ABT_eventual` use similar but different code. `ABT_mutex` uses an incorrect queue implementation (see #14, #41, and #102).

## Solution

To avoid code duplication, this PR implements `ABTI_waitlist`.  `ABT_mutex` will use this when `--disable-simple-mutex` is passed; by default the existing yield-based implementation is used. Better `ABT_mutex` implementation is our future work.

### Expected impact

This change significantly affects the performance of `ABT_mutex` with `--disable-simple-mutex`. The new implementation is correct but slower in some cases. This PR should not affect the performance of the other synchronization objects (including `ABT_mutex` without `--disable-simple-mutex`) much -- this PR should make them faster.


shintaro-iwasaki:
test:argobots/all
test:argobots/freebsd
test:argobots/osx
test:argobots/solaris

PR #270, opened 2020-12-03
stream_barrier: add a fallback implementation of ABT_stream_barrier

shintaro-iwasaki:
`ABT_stream_barrier` was unavailable if `pthread_barrier` is not supported. Although such a case should be extremely rare, we should provide the same functionality if it should happen.  This patch implements a simple sense-based barrier implementation (though it uses `uint64`-type tag instead of `bool`-type sense).  See https://en.wikipedia.org/wiki/Barrier_(computer_science) for details.

Note that `ABT_stream_barrier` uses `pthread_barrier` by default, so this PR will not affect execution on almost all platforms.

### DIFF
--- a/src/include/abti.h
+++ b/src/include/abti.h
@@ -415,7 +415,13 @@ struct ABTI_barrier {
 
 struct ABTI_xstream_barrier {
     uint32_t num_waiters;
+#ifdef HAVE_PTHREAD_BARRIER_INIT
     ABTD_xstream_barrier bar;
+#else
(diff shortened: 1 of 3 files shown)


shintaro-iwasaki:
test:argobots/all

shintaro-iwasaki:
All the tests passed when `pthread_barrier` is manually turned off.

shintaro-iwasaki:
test:argobots/all

PR #271, opened 2020-12-08
future: call a callback function before making a future ready

shintaro-iwasaki:
## Problem

```c
void thread1(void *) {
  while (1) {
    ABT_bool flag;
    ABT_future_test(future, &flag);
    if (flag == ABT_TRUE)
      printf("future is ready!\n");
  }
}

void thread2(void *) {
  ABT_future_set(future, NULL);
}
```

`ABT_future_test()` could return `ABT_TRUE` before `ABT_future_set()` calls a callback function. Though this behavior is not mentioned in the Argobots' document, the callback function should be called before the future becomes ready so that waiters can depend on execution done by the callback function.

## Solution
This patch guarantees that a callback function is called before a future gets ready.

## Impact
No performance impact is expected. This change should not affect the behavior of existing applications.

### DIFF
--- a/src/futures.c
+++ b/src/futures.c
@@ -189,19 +189,23 @@ int ABT_future_set(ABT_future future, void *value)
     ABTI_spinlock_acquire(&p_future->lock);
 
     size_t counter = ABTD_atomic_relaxed_load_size(&p_future->counter);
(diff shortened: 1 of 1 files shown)


shintaro-iwasaki:
test:argobots/all

PR #275, opened 2020-12-22
tool: API clean up

shintaro-iwasaki:
This PR updates the specification of the current tool API for the upcoming Argobots 1.1

Specifically, this PR merges the two versions of tool functions for a ULT and a tasklet into one to make the API simpler. The user should use `ABT_tool_register_thread_callback` for tasklet profiling.

The tool interface is not included in Argobots 1.0, so this change should not break any API and ABI compatibility. By default the tool interface is disabled, so it will not affect the performance of most applications.


shintaro-iwasaki:
test:argobots/all

PR #276, opened 2020-12-22
info: improve API

shintaro-iwasaki:
This PR updates the specification of the current info API for the upcoming Argobots 1.1.

Specifically, this PR restores the old error checks that existed in Argobots 1.0 for the backward compatibility. This PR itself does not hurt the performance. Unless the user program highly relies on the error code (not an error, an "error code"), this PR does not affect the behavior of the program.


shintaro-iwasaki:
test:argobots/all

PR #278, opened 2020-12-23
synchronization objects: improve API

shintaro-iwasaki:
This PR improves the API of synchronization objects: `ABT_mutex` (and `ABT_mutex_attr`), `ABT_future`, `ABT_barrier`, `ABT_xstream_barrier`, `ABT_eventual`, `ABT_cond`, and `ABT_rwlock`.

This PR changes the API of Argobots 1.0, which we believe is a bug in the original document (or the implementation), although most part of the Argobots 1.0 API will be preserved (even though it does not make much sense).

- `ABT_future`: the order of compartments gets "undefined".
  Originally, the document says this order follows the call order of `ABT_future_set()`, but the user cannot observe which caller calls first if they are called simultaneously.   There is no point of strictly defining the order, so this order becomes undefined from Argobots 1.1.

- `ABT_eventual_set()`: setting a value to a ready eventual gets an error.
  Previously, the implementation allows setting a value to a ready eventual. From the semantics of `ABT_eventual` (which is similar to `std::future`), setting a value more than once should not be allowed. Argobots 1.1 makes it an error. Note that this readiness is updated atomically, so `ABT_eventual_set()` guarantees the atomicity of the value update (and also only one updater can update it and get `ABT_SUCCESS`).

- `ABT_mutex_create_with_attr()`: passing `ABT_MUTEX_ATTR_NULL` to this function creates a default mutex
  Previously, passing `ABT_MUTEX_ATTR_NULL` to `ABT_mutex_create_with_attr()` was an error, but most Argobots functions allow `ABT_XXX_ATTR_NULL` or `ABT_XXX_CONFIG_NULL`.  To fix this inconsistency, Argobots 1.1. allows `ABT_MUTEX_ATTR_NULL`  for `ABT_mutex_create_with_attr()`.

- `ABT_barrier_create()`, `ABT_barrier_reinit()`: "0-num-waiters" barrier is forbidden (and thus an error is returned).
   A barrier that does not have a waiter is nonsense (waiting on such a barrier itself is an error??). When the user pass "0" as `num_waiters`, those functions now return an error.

I believe most programs do not rely on the Argobots behavior in such a corner case, but if this change breaks your code, please let us know so that we can revert this change.

Note that Argobots 2.0 which should come after Argobots 1.1 will more aggressively change the API.


shintaro-iwasaki:
test:argobots/all

PR #279, opened 2020-12-23
self: improve API

shintaro-iwasaki:
This PR improves the API and its explanation of `ABT_self_xxx()` functions for the upcoming Argobots 1.1.

This PR does not change the API of Argobots 1.0 except for the following "bug".

* `ABT_self_suspend()`: it returns `ABT_ERR_INV_XSTREAM` if it is called by an external thread.

  In Argobots 1.0, it returns `ABT_ERR_INV_THREAD` in such a case. I believe it is a bug, not an intended error code (for some particular reasons). Note that all the functions in Argobots return `ABT_ERR_INV_XSTREAM` when an external thread wrongly calls a routine.

Because Argobots-1.0 compatible error check is slightly heavier, some `ABT_self_xxx()` functions might be slightly slower (say, five cycles or so), but overall I don't think this change affect any program.

This PR also replaces "main ULT" with "primary ULT" since "primary ULT" is the correct term in Argobots.

### DIFF
--- a/Doxyfile.in
+++ b/Doxyfile.in
@@ -1822,7 +1822,11 @@ ALIASES += DOC_CONTEXT_CTXSWITCH="This routine may perform context switching for
 
 ALIASES += DOC_CONTEXT_INIT="This routine can be called in any execution context. Argobots must be initialized."
 
+ALIASES += DOC_CONTEXT_INIT_NOEXT="This routine can be called in either a ULT context or a tasklet context. Argobots must be initialized."
+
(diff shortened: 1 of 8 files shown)


shintaro-iwasaki:
test:argobots/all

shintaro-iwasaki:
test:argobots/all

shintaro-iwasaki:
test:argobots/all

shintaro-iwasaki:
test:argobots/all

PR #281, opened 2020-12-24
sched: improve API

shintaro-iwasaki:
This PR improves the API and its explanation of `ABT_sched_xxx()` (including `ABT_sched_config_xxx()`) functions for the upcoming Argobots 1.1.

This PR does not change the API of Argobots 1.0 except for the following "behavior".

- `ABT_sched_type`: regardless of this value, the runtime will always create a ULT-type scheduler. A tasklet-type scheduler is too restrictive and the runtime cannot support it. Note that this change has been already adopted by #182.

- `ABT_sched_config_access`: regardless of this value, the runtime will always create a MPMC-type pool for a `ABT_POOL_NULL` pool. This is because non-MPMC pools are too restrictive (e.g., regarding a resume operation). The user can manually create such a pool by `ABT_pool_create_basic()`.

This PR also fixes some bugs (e.g.,  memory leak on an error path).

### DIFF
--- a/Doxyfile.in
+++ b/Doxyfile.in
@@ -1842,15 +1842,21 @@ ALIASES += DOC_DESC_ATOMICITY_EVENTUAL_READINESS="The readiness of an eventual i
 
 ALIASES += DOC_DESC_ATOMICITY_FUTURE_READINESS="The readiness of a future is managed atomically."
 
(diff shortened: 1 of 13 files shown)


shintaro-iwasaki:
test:argobots/all

shintaro-iwasaki:
test:argobots/all

shintaro-iwasaki:
test:argobots/all

shintaro-iwasaki:
test:argobots/all

PR #282, opened 2020-12-28
pool: improve API

shintaro-iwasaki:
This PR improves the API and its explanation of `ABT_pool_xxx()` functions for the upcoming Argobots 1.1.

The changes of this PR are relatively larger than the others, but this PR does not affect the behavior of a program (much) if a program does not use user-defined pools. In other words, unless your program uses `ABT_pool_create()` (*not `ABT_pool_create_basic()`*), these API changes are almost negligible.

### User-defined pools
- `ABT_pool_access`: the Argobots runtime uses access only to determine the pool implementation.

  It has been already removed by #239, but this change will be applied to Argobots 1.1. The user needs to be responsible for pool access management.
- `ABT_pool_def`: `u_get_type()`, `u_create_from_task()`, and `ABT_unit_get_task_fn` are unused. Instead, `u_create_from_thread()` is used for tasklets.

  All units including both ULTs and tasklets are created by `u_create_from_thread()`. This makes the implementation of the user-defined pool much easier since the user does not need to return a type of a work unit from `ABT_unit`, which requires a hash table.

### Others
- `ABT_pool_pop()`, `ABT_pool_pop_timedwait()`, and `ABT_pool_remove()`: calling these functions on an external thread is allowed.

  There is no reason to prohibit it. Argobots 1.1 allows it.

- `ABT_pool_push()`: `ABT_ERR_INV_UNIT` is returned instead of `ABT_ERR_UNIT` if `ABT_UNIT_NULL` is passed.

  This change makes a returned error more consistent.

### DIFF
--- a/Doxyfile.in
+++ b/Doxyfile.in
@@ -1856,6 +1856,8 @@ ALIASES += DOC_DESC_V10_ACCEPT_TASK{1}="\DOC_V10 The user cannot pass a tasklet 
 
 ALIASES += DOC_DESC_V10_ACCEPT_THREAD{1}="\DOC_V10 The user cannot pass a ULT handle as \1.\n \DOC_V11 This routine accepts a ULT handle as \1.\n @rationale Argobots 2.0 integrates a ULT and a tasklet into a single thread concept to make the API more general.  Argobots 1.1 has introduced this change since this change does not break the compatibility of API and ABI of Argobots 1.0. @endrationale"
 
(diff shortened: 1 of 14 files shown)


shintaro-iwasaki:
test:argobots/all

shintaro-iwasaki:
test:argobots/all

shintaro-iwasaki:
test:argobots/all

shintaro-iwasaki:
test:argobots/all

PR #283, opened 2020-12-29
xstream: improve API 

shintaro-iwasaki:
This PR improves the API and its explanation of `ABT_xstream_xxx()` functions for the upcoming Argobots 1.1.

This PR does not change the API of Argobots 1.0 except for the following.

- `ABT_xstream_free()`: returns `ABT_ERR_INV_XSTREAM` instead of `ABT_SUCCESS` when the user tries to free `ABT_XSTREAM_NULL`.

  Freeing `ABT_XSTREAM_NULL` is an obvious error. `ABT_SUCCESS` should not be returned.

- `ABT_xstream_set_main_sched()`: allows changing a scheduler even if it has an empty pool.

  There is no reason to return an error when a scheduler has an empty pool. It is the user's responsibility to schedule work units in a pool associated with the old scheduler.

This PR also fixes some bugs (e.g., `ABT_xstream_set_rank()`).

### DIFF
--- a/Doxyfile.in
+++ b/Doxyfile.in
@@ -1826,6 +1826,8 @@ ALIASES += DOC_CONTEXT_INIT_NOEXT="This routine can be called in either a ULT co
 
 ALIASES += DOC_CONTEXT_INIT_NOTASK="This routine can be called in either a ULT context or an external thread context. Argobots must be initialized."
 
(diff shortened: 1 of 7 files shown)


shintaro-iwasaki:
test:argobots/all

shintaro-iwasaki:
test:argobots/all

shintaro-iwasaki:
test:argobots/all

PR #284, opened 2020-12-30
thread: improve API

shintaro-iwasaki:
This PR improves the API and its explanation of `ABT_thread_xxx()` and `ABT_thread_attr_xxx()` functions for the upcoming Argobots 1.1.

This PR does not change the API of Argobots 1.0 except for the following.

- `ABT_thread_migrate_xxx()`: removes error checks for target threads/schedulers/pools/execution streams.

  Since many instances are involved in the Argobots migration mechanism, it is practically impossible to atomically perform migration operations. Hence, the migration (implicitly) needed the user to guarantee that all targets are alive (i.e., not freed/terminated) during the migration process. Previous implementations of `ABT_thread_migrate_xxx()` checked some of them, but these checks are removed since the specification explicitly mentions that it is the user's responsibility.

  If the program does not rely on an error returned by `ABT_thread_migrate_xxx()`, this change should not affect the behavior.

- `ABT_thread_attr_set_stack()`: returns `ABT_ERR_INV_ARG` instead of `ABT_ERR_OTHERS` if the stack size is properly aligned.

  This change is for consistent error code.

In addition, most `ABT_thread_xxx()` functions now accept a tasklet as an input.

### DIFF
--- a/Doxyfile.in
+++ b/Doxyfile.in
@@ -1954,6 +1954,8 @@ ALIASES += DOC_ERROR_INV_ARG_GREATER_THAN{2}="If \1 is larger than \2, \c ABT_ER
 
 ALIASES += DOC_ERROR_INV_ARG_INV_SCHED_PREDEF{1}="If \1 is not a valid predefined scheduler type, \c ABT_ERR_INV_ARG is returned.\n"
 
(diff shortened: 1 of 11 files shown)


shintaro-iwasaki:
test:argobots/all

shintaro-iwasaki:
test:argobots/all

Issue #287, opened 2021-01-06
synchronization: guarantee that a ULT does not yield if no contention 


shintaro-iwasaki:
This issue is raised by @carns. https://lists.argobots.org/pipermail/discuss/2021-January/000094.html

We need to examine the implementation, but in any case the Argobots specification should guarantee that a ULT that locks/unlocks asynchronization object may not spuriously yield if there is no contention (e.g., `mutex`) and otherwise explicitly stated. At least I implemented so as a developer, but this behavior should be clarified and tested. Maybe there are some bugs.

carns:
Just to follow up on the github issue: the only instance I have observed where this guarantee does not hold is on a power9 CPU with Argobots 1.0.  It works as expected now with origin/main.

PR #288, opened 2021-01-08
sync: ULTs should not yield if there is no contention

shintaro-iwasaki:
This PR addresses the issue #287 from both documentation and implementation perspectives.
A ULT may yield when it tries to take a lock but that lock is already locked by another work unit, but otherwise a ULT should not yield -- this behavior is natural and intuitive. Since some users rely on this behavior, we should clarify this behavior in the function description. This PR also adds a test to examine this behavior.

At this point, this patch itself does not change the existing Argobots implementation since it satisfies this behavior.


shintaro-iwasaki:
test:argobots/all

shintaro-iwasaki:
test:argobots/all

Issue #274, opened 2020-12-21
ULT stack allocation method to address overrun scenario


bfaccini:
We regularly trigger crashes due to ULT stack overrun, mainly caused by 3rd party libraries who are stacking big arrays/structs, or also during lazy symbol resolution of dynamic libs...

Since we expect to run Millions of concurrent ULTs we could not rely on statically allocating big chunk for stacks to limit memory footprint and thus we would like to benefit of the existing kernel mechanism which allows stacks to dynamically raise to their associated limit if needed.

We first thought that we could implement this externally from Argobots and use our own external stacks in Argobots, but as it is described in issue #16 the present lack of hook in Argobots that would allow to be aware that a stack is no longer in use and be able to free/reassign it, prevents it.

So, here are the main point that should be addressed in order to implement this new stack allocation mechanism inside Argobots: 
                  _ allocate a new stack using mmap(), with a minimal size and by ensuring it will be positioned in a hole (by providing its address as a hint to mmap()!) far enough from the preceding mapping, this to be allowed to automatically grow downward (as per the Kernel's existing mechanism for stacks!), until some reasonable minimum size (assuming that preceding mapping will not itself grow upward...) which could be set as the xstream's stack limit.

                   _ as  it is not possible to rely on kernel to find the best place for us (ie by using NULL as mmap()'s address/1st parameter), we will have to parse /proc/self/maps current content and decide. This process will also have to take into account of current/previous stacks allocations in order to avoid re-using a hole where a previously allocated stack is already expected to grow...

                   _ as part of these required mechanisms need for managing stacks, a map of current allocations has to be maintained to avoid possibly colliding allocation of new stack, and also a mechanism/policy to determine no longer used stacks to be re-allocated or freed/munmap()'ed.

                   _ also, Kernel/boot "stack_guard_gap" parameter will have to be taken into account, as the minimum hole size between the preceding mapping, because it is used by Kernel to decide if a "stack" region is allowed to grow or not.

carns:
I'm not familiar enough with these mechanism to comment on @bfaccini 's proposal, but I am interested in this topic.

[...]

PR #293, opened 2021-01-27
mem: implement a stack overflow detection mechanism

shintaro-iwasaki:
## Problems

With Argobots, stack overflow can easily happen because each stack tends to be small (e.g., #274), but this overflow is often silent because there is no stack guard page. Creating an `mprotect()`-ed stack is not very trivial, but at least Argobots should employ a simple mechanism to detect it.

## Solution

This patch implements a stack canary that can check whether a stack overflow happens or not when a ULT is freed. This check is quite simple and lightweight. If stack overflow is detected, the runtime terminates. Please check the mechanism here: https://en.wikipedia.org/wiki/Stack_buffer_overflow#Stack_canaries

Note that this mechanism is to make sure that there is no stack overflow. This mechanism cannot pinpoint where the stack overflow happens.

This mechanism slightly degrades the performance, so it is disabled by default. It can be enabled by setting `--enable-stack-overflow-check=canary` (or `--enable-stack-overflow-check=canary-XX` where XX is a large number to use a larger canary) at configuration time.

## Impact

This feature is disabled by default, so it will not affect the default behavior.


shintaro-iwasaki:
test:argobots/all
test:argobots/freebsd
test:argobots/solaris

shintaro-iwasaki:
test:argobots/all

shintaro-iwasaki:
test:argobots/all

shintaro-iwasaki:
test:argobots/all

PR #295, opened 2021-01-28
configure.ac: support autoconf 2.70

shintaro-iwasaki:
This PR updates `configure.ac` to suppress warnings newly issued by autoconf 2.70.

Note that this patch does not change the autoconf version requirement. We will continue to support autoconf 2.67, 2.68, and 2.69.
This PR also removes unnecessary macros from `*.m4` to make the maintenance easy.


shintaro-iwasaki:
test:argobots/all
test:argobots/osx
test:argobots/freebsd
test:argobots/solaris

PR #296, opened 2021-01-29
mutex: add new functions to get a recursive property of a mutex

shintaro-iwasaki:
This PR adds two new functions, `ABT_mutex_get_attr()` and `ABT_mutex_attr_get_recursive()`, which are useful to get a recursive property of a mutex.

They are implemented for completeness of the API, but these functions are useful for debugging too.


shintaro-iwasaki:
test:argobots/all

Issue #145, opened 2020-02-19
Implement static initializers of synchronization objects. 


shintaro-iwasaki:
Argobots functions including `ABT_mutex_create()` etc need to be called *after* `ABT_init()`, so it is cumbersome to initialize a global variable. The typical use case is a global lock that exists only in a local translation unit. 
```c
// foo.c
static ABT_mutex g_foo_lock;
void runtime_foo_func() {
  ABT_mutex_lock(&g_foo_lock);
  // a critical section protected by g_foo_lock
  ABT_mutex_unlock(&g_foo_lock);
}

// main.c
void runtime_init() {
  ABT_init(0, NULL);
  // how to refer to g_foo_lock for initialization?
}
void runtime_fini() {
  // how to finalize g_foo_lock?
  ABT_finalize(0, NULL);
}
```
It is actually required in the context of Open MPI + Argobots (https://github.com/open-mpi/ompi/pull/6578). We should provide a static initializer (e.g., `PTHREAD_MUTEX_INITIALIZER`).

shintaro-iwasaki:
Though Pthreads provides static initializers for barrier and rwlock, I believe most users need a static initializer for mutex and condition variable. Both `ABT_mutex` and `ABT_cond` are now statically initialized by #299 and #300, so we can close this issue.

This static initialization is also beneficial to reduce "ABT_xxx_create" overheads. Please create an issue or a PR if you want Argobots to cover synchronization objects other than `ABT_mutex` and `ABT_cond`.

PR #299, opened 2021-02-05
mutex: mutex without Argobots initialization 

shintaro-iwasaki:
## Pull Request Description
Argobots assumes that the Argobots functionality is available only between `ABT_init()` and `ABT_finalize()`, so `ABT_mutex` can be created and used only when Argobots is initialized, but this makes it difficult to port existing Pthreads-based applications and runtimes if they do not have explicit initialization and finalization phases.  For example, some locks used in BOLT (Argobots-aware LLVM OpenMP) and Argobots-aware Open MPI can be called before `ABT_init()`, so currently those locks are implemented by combining `ABT_initialized()`, `ABT_thread_yield()`, and some atomic operations.  This is very cumbersome.

This PR provides a method to use an Argobots mutex even if Argobots is not initialized.  `ABT_mutex_memory` can be statically initialized.  Lock and unlock functions such as `ABT_mutex_lock()`, `ABT_mutex_unlock()`, and `ABT_mutex_trylock()` now work without Argobots initialization, so users can use those lock and unlock functions with `ABT_mutex` converted from `ABT_mutex_memory` before Argobots initialization.  If external threads are supported (this is enabled by default), such an `ABT_mutex` can always serialize concurrent accesses from external threads (=Pthreads) and ULTs.

```c
// ABT_RECURSIVE_MUTEX_INITIALIZER is also available.
ABT_mutex_memory mutex_mem = ABT_MUTEX_INITIALIZER;
int g_protected_value = 0;
void inc_protected_value()
{
    ABT_mutex mutex = ABT_MUTEX_MEMORY_GET_HANDLE(&mutex_mem);
    // inc_protected_value() can be used for Pthreads without ABT_init()
    // while it can be used for ULTs as well as Pthreads after ABT_init().
    ABT_mutex_lock(mutex);
    g_protected_value++;
    ABT_mutex_unlock(mutex);
}
````

This PR should not degrade the performance of the existing mutex implementation (~especially if `--enable-feature=no-ext-thread` is not explicitly set by users~  *[EDIT] there is no performance difference even if `--enable-feature=no-ext-thread` is set, but `ABT_mutex_xxx()` cannot be called without `ABT_init()` since the execution context is an external thread*).

This is not the main goal of this PR, but using `ABT_mutex_memory` is more lightweight than `ABT_mutex_create()`, so using this static initializer should be beneficial if applications create many `ABT_mutex`.

## Checklist

### DIFF
--- a/src/include/abt.h.in
+++ b/src/include/abt.h.in
@@ -1072,6 +1072,95 @@ typedef enum ABT_sync_event_type            ABT_sync_event_type;
 #endif
 
 /**
+ * @ingroup MUTEX
+ * @brief   A struct that can be converted to ABT_mutex.
+ * @hideinitializer
+ *
+ * \c ABT_mutex_memory is a struct to statically allocate and initialize
+ * \c ABT_mutex.  \c ABT_mutex_memory can be statically initialized by
+ * \c ABT_MUTEX_MEMORY_INITIALIZER and \c ABT_RECURSIVE_MUTEX_INITIALIZER and
+ * can be converted to \c ABT_mutex by \c ABT_MUTEX_MEMORY_GET_HANDLE().
+ *
+ * \c dummy may not be accessed by a user.
+ */
+typedef struct {
+    int dummy[16];
+} ABT_mutex_memory;
+
+/**
+ * @ingroup MUTEX
+ * @brief   Initialize \c ABT_mutex_memory.
+ *
+ * \c ABT_MUTEX_INITIALIZER() statically initializes \c ABT_mutex_memory.  The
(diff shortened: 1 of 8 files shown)


shintaro-iwasaki:
test:argobots/all
test:argobots/freebsd
test:argobots/osx
test:argobots/solaris

shintaro-iwasaki:
This PR partially solves #145 while this PR only implements a mutex.

shintaro-iwasaki:
test:argobots/all
test:argobots/freebsd
test:argobots/osx
test:argobots/solaris

PR #306, opened 2021-03-01
efficient synchronization of external threads using futex

shintaro-iwasaki:
### Problem

Most Argobots functions now support external threads; Pthreads can call Argobots synchronization functions such as `ABT_mutex_lock()`, `ABT_barrier_wait()`, and `ABT_thread_join()` together with ULTs.  However, the current implementation of Argobots uses spin-wait (maybe plus `pause` or `sched_yield()`), which burns cores if synchronization conditions are not satisfied soon.  This can cause a catastrophic performance degradation in cases where Pthreads would call such synchronization functions, for example, when users do not know who will call Argobots functions.  This issue can happen particularly when Argobots is used as a building block of another runtime system that is expected to be called by other high-level programs.

### Solution

**This change affects only cases where Pthreads (neither ULTs nor tasklets!) directly call Argobots synchronization functions.**

Argobots provides a futex-based suspension configuration that affects all the synchronization operations.  Currently, unless the user passes `--enable-wait-policy=active`, Argobots will make the underlying external thread block on a futex (like most POSIX synchronization operations internally do).  The underlying external thread will not use CPU resource in a busy loop, so it alleviates thread oversubscription.  This does not change the ULT performance, so if you call such synchronization objects on ULTs (this is a typical use case), this change does not affect the behavior (i.e., it does not increase the overheads of these operations).

### Developer Notes

This synchronization mechanism is a little bit complicated.  Just some notes:

1. futex is basically Linux only.  As a fallback, Argobots has a POSIX-based implementation (`pthread_mutex_t` and `pthread_cond_t`) for other UNIX systems.

2. Since we should not allocate large memory space just for this additional mechanism that not many programs use, `ABTD_futex_xxx_t` should be small and simple.  Currently, its size is 4 or 8 bytes.  This structure can be initialized by zero clear no matter whether POSIX or futex is used (note: `pthread_mutex_t` or `pthread_cond_t` can be 32 bytes or more (implementation dependent)).  This is also necessary to coexist this feature with static initializers of `ABT_mutex` and `ABT_cond`.

3. The synchronization semantics of some synchronization operations is different.  Specifically, `ABT_cond_wait()` does not allow spurious wakeup while `pthread_cond_wait()` allows it.  This causes a semantics mismatch and adds some overhead.

4. External signal can disrupt system calls (i.e., futex or `pthread_cond_wait`).  Seven tests are newly added to check if external threads work while receiving a lot of signals (`ext_thread_join`, `ext_thread_mutex`, ...).

### Performance

The following shows the performance of synchronization objects of the following:
 1. corresponding POSIX implementation (`pthread_mutex_lock()` etc)
 2. this PR with futex (I tried this on Linux so futex was available on all the machines)
 3. this PR without futex (for reference)
 4. this PR with the active wait policy (=the current Argobots behavior)

on those machines:

 1. Intel Skylake (Intel Xeon 8180M, 28 cores / 56 hardware threads * 2 sockets) / openSUSE 42.2 / GCC 7.5
 2. Summit-like POWER 9 (IBM S822LC 10 cores / 80 hardware threads * 2 sockets) / RedHat 7.6 / GCC 9.3
 3. 64-bit ARM (JLSE's 64-bit ARM for HPC, in total 112 hardware threads)/ RedHat 7.6 / GCC 9.3

All use Pthreads for underlying threads.

<details><summary>(The benchmark code is collapsed)</summary>
<p>

```c
/* gcc -O3 test.c -lpthread */
#include <abt.h>
#include <pthread.h>
#include <stdio.h>
#include <stdlib.h>
#include <assert.h>

int g_cond_val = 0;
pthread_barrier_t pth_barrier;
pthread_mutex_t pth_mutex;
pthread_cond_t pth_cond;

ABT_barrier abt_barrier;
ABT_mutex abt_mutex;
ABT_cond abt_cond;

int g_num_pthreads = 0;
int g_num_repeats = 0;
int g_type = 0;

typedef struct {
    int tid;
    pthread_t thread;
} pthread_arg_t;

void *pthread_func(void *arg)
{
    pthread_arg_t *p_arg = (pthread_arg_t *)arg;
    int tid = p_arg->tid;
    int num_pthreads = g_num_pthreads;
    int num_repeats = g_num_repeats;

    pthread_barrier_wait(&pth_barrier);
    g_cond_val = 0;
    double start_time = ABT_get_wtime();
    pthread_barrier_wait(&pth_barrier);

    if (g_type == 0) {
        /* Pthreads mutex */
        for (int i = 0; i < num_repeats; i++) {
            pthread_mutex_lock(&pth_mutex);
            /* Do nothing. */
            pthread_mutex_unlock(&pth_mutex);
        }
    } else if (g_type == 1) {
        /* Argobots mutex */
        for (int i = 0; i < num_repeats; i++) {
            ABT_mutex_lock(abt_mutex);
            /* Do nothing. */
            ABT_mutex_unlock(abt_mutex);
        }
    } else if (g_type == 2) {
        /* Pthreads cond broadcast */
        assert(g_num_pthreads >= 2);
        while (g_cond_val < num_repeats) {
            pthread_mutex_lock(&pth_mutex);
            if (g_cond_val % num_pthreads == tid) {
                g_cond_val += 1;
                pthread_cond_broadcast(&pth_cond);
            } else if (g_cond_val < num_repeats) {
                pthread_cond_wait(&pth_cond, &pth_mutex);
            }
            pthread_mutex_unlock(&pth_mutex);
        }
    } else if (g_type == 3) {
        /* Argobots cond broadcast */
        assert(g_num_pthreads >= 2);
        while (g_cond_val < num_repeats) {
            ABT_mutex_lock(abt_mutex);
            if (g_cond_val % num_pthreads == tid) {
                g_cond_val += 1;
                ABT_cond_broadcast(abt_cond);
            } else if (g_cond_val < num_repeats) {
                ABT_cond_wait(abt_cond, abt_mutex);
            }
            ABT_mutex_unlock(abt_mutex);
        }
    } else if (g_type == 4) {
        /* Pthreads barrier */
        for (int i = 0; i < num_repeats; i++)
            pthread_barrier_wait(&pth_barrier);
    } else if (g_type == 5) {
        /* Argobots barrier */
        for (int i = 0; i < num_repeats; i++)
            ABT_barrier_wait(abt_barrier);
    }
    pthread_barrier_wait(&pth_barrier);

    if (tid == 0) {
        double elapsed_time = ABT_get_wtime() - start_time;
        printf("Unit execution time: %.6f [us]\n",
               elapsed_time / num_repeats * 1.0e6);
    }
    return NULL;
}

{
    if (argc != 4) {
        printf("Usage: ./a.out NUM_THREADS NUM_REPEATS BENCH_TYPE\n");
        printf("BENCH_TYPE = 0: POSIX mutex\n");
        printf("           = 1: Argobots mutex\n");
        printf("           = 2: POSIX cond\n");
        printf("           = 3: Argobots cond\n");
        printf("           = 4: POSIX barrier\n");
        printf("           = 5: Argobots barrier\n");
        return -1;
    }
    g_num_pthreads = atoi(argv[1]);
    g_num_repeats = atoi(argv[2]);
    g_type = atoi(argv[3]);

    /* Performance check. */
    ABT_init(0, NULL);

    /* Allocate all synchronization objects. */
    pthread_barrier_init(&pth_barrier, NULL, g_num_pthreads);
    pthread_mutex_init(&pth_mutex, NULL);
    pthread_cond_init(&pth_cond, NULL);
    ABT_barrier_create(g_num_pthreads, &abt_barrier);
    ABT_mutex_create(&abt_mutex);
    ABT_cond_create(&abt_cond);

    pthread_arg_t *pthread_args =
        (pthread_arg_t *)malloc(sizeof(pthread_arg_t) * g_num_pthreads);

    /* Check the mutex performance.  Note that we do not consider a fairness
     * issue */
    for (int i = 0; i < g_num_pthreads; i++) {
        pthread_args[i].tid = i;
        pthread_create(&pthread_args[i].thread, NULL, pthread_func,
                       &pthread_args[i]);
    }
    for (int i = 0; i < g_num_pthreads; i++) {
        pthread_join(pthread_args[i].thread, NULL);
    }

    pthread_barrier_destroy(&pth_barrier);
    pthread_mutex_destroy(&pth_mutex);
    pthread_cond_destroy(&pth_cond);
    ABT_barrier_free(&abt_barrier);
    ABT_mutex_free(&abt_mutex);
    ABT_cond_free(&abt_cond);
    free(pthread_args);

    ABT_finalize();
    return 0;
}
```

<p>
</details>

<details><summary>(The experimental setting is collapsed)</summary>
<p>

Little contention:
Mutex: 1 thread (Skylake/ARM64/POWER9)
Cond: 2 threads (Skylake/ARM64/POWER9)
Barrier: 2 threads (Skylake/ARM64/POWER9)

High contention (not oversubscribed):
Mutex/Cond/Barrier: 28 threads (Skylake/ARM64), 40 threads (POWER9)

High contention (oversubscribed):
Mutex/Cond/Barrier: 256 threads (Skylake/ARM64), 320 threads (POWER9)

Results are the average of 6 time executions.  Each execution repeats these operations many times (around 100000 or more).  Argobots is compiled with `--enable-perf-opt`.

<p>
</details>

image

Please note that this result is to "understand" the performance.  This PR's change is qualitative, so bad performance of microbenchmarks does not lessen the value of this PR much.  This microbenchmark does not well represent a use case of Pthreads+Argobots: the purpose of this new mechanism is to sleep external threads while waiting for others' work that is moderately coarse-grained and if it is the case this PR works very well no matter POSIX or futex is used.  I also note that semantics (spurious wakeup or not), provided features (Pthreads only or ULT + Pthreads), and scheduling fairness (Argobots synchronization objects might not be as fair as those of Pthreads) are different, so this comparison is not fair.

We can observe the following:
1. In any case, this mechanism works on several architectures without hang.
2. If not contended, the performance of Argobots looks okay.
3. If contended, the performance of Argobots is not good.
4. If the wait policy is passive, overall the futex version is faster than the Pthreads version.
5. When Pthreads oversubscription happens, the performance of `ABT_barrier` and `ABT_cond` gets significantly bad when the wait policy of Argobots is active (because Argobots spins cores).

There is much room for improvement, so if further performance optimization is necessary, please let us know.

### Known issues

1. Some operations are obviously unoptimized (specifically `ABT_cond_signal()`).
2. A few operations still use busy-wait (as far as I checked, when a thread waits for completion of a tasklet).
3. Some users might want to set this configuration per synchronization objects.  Perhaps users wants to change this behavior per function (e.g., `ABT_mutex_lock_active()` and `ABT_mutex_lock_passive()`).  This is future work (if there is such a demand).

## Checklist

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -207,6 +207,14 @@ AC_ARG_ENABLE([stack-overflow-check],
         none|no
 ],,[enable_stack_overflow_check=no])
 
+# --enable-wait-policy
+AC_ARG_ENABLE([wait-policy],
+[  --enable-wait-policy@<:@=OPTS@:>@ set a wait policy of blocking operations
+        default|auto - choose the best policy, which is currently "passive".
+        active       - spin-wait on blocking.  It uses more CPU resources.
+        passive      - suspend CPU cores on blocking.  It increases latency.
+],,[enable_wait_policy=auto])
+
 # --disable-simple-mutex
 AC_ARG_ENABLE([simple-mutex],
     AS_HELP_STRING([--disable-simple-mutex], [use an experimental mutex implementation]),,
@@ -774,6 +782,10 @@ AS_IF([test "x$have_linux_futex" = "xyes"],
       [AC_DEFINE(ABT_CONFIG_USE_LINUX_FUTEX, 1,
                  [Define to use Linux-type futex])])
 
+# --enable-wait-policy
+AS_IF([test "x$enable_wait_policy" = "xactive"],
+      [AC_DEFINE(ABT_CONFIG_ACTIVE_WAIT_POLICY, 1,
+                 [Define to enable the active wait policy])])
 
 # --disable-simple-mutex
 AS_IF([test "x$enable_simple_mutex" = "xyes"],
--- a/configure.ac
+++ b/configure.ac
@@ -875,6 +875,16 @@ AC_CHECK_LIB(pthread, pthread_join)
 # check pthread_barrier
 AC_CHECK_FUNCS(pthread_barrier_init)
 
+# check -lrt library
+AC_CHECK_LIB(rt, timer_create)
+if test "$ac_cv_lib_rt" = "yes" ; then
+    # -lrt is used for testing
+    abttest_timer_library="-lrt"
+else
+    abttest_timer_library=""
+fi
+AC_SUBST(abttest_timer_library)
+
 # check timer functions
 AC_CHECK_FUNCS(clock_gettime mach_absolute_time gettimeofday)
 if test "$ac_cv_func_clock_gettime" = "yes" ; then
--- a/configure.ac
+++ b/configure.ac
@@ -747,6 +747,34 @@ if test "$CC" != "xlc" -a "$CC" != "pgcc" -a "x$have_atomic_builtin" = "xyes" ; 
 fi
 
 
+# check if futex is available.  Note that futex() is basically Linux-specific.
+AC_COMPILE_IFELSE(
+[AC_LANG_PROGRAM([
+ #include <unistd.h>
+ #include <linux/futex.h>
+ #include <syscall.h>
+ #include <assert.h>
+],[
+ // Maybe we should not include <futex.h> since it seems that it has some
+ // compatibility issues (see LLVM OpenMP).  We should use a constant value
+ // for futex_op if we found some compilation issues here.
+ int ret, futex_val = 0;
+ // The following does not sleep since futex_val is not 1.
+ ret = syscall(SYS_futex, &futex_val, FUTEX_WAIT_PRIVATE, 1, 0, 0, 0);
+ assert(ret == 0);
+ // The following does not wake up any since no thread is sleeping.
+ ret = syscall(SYS_futex, &futex_val, FUTEX_WAKE_PRIVATE, 1, 0, 0, 0);
+ // The return value is 0 since no thread was waken up.
+ assert(ret == 0);
+])],
+[have_linux_futex=yes],
+[have_linux_futex=no]
+)
+AS_IF([test "x$have_linux_futex" = "xyes"],
+      [AC_DEFINE(ABT_CONFIG_USE_LINUX_FUTEX, 1,
+                 [Define to use Linux-type futex])])
+
+
 # --disable-simple-mutex
 AS_IF([test "x$enable_simple_mutex" = "xyes"],
       [AC_DEFINE(ABT_CONFIG_USE_SIMPLE_MUTEX, 1,
--- a/src/arch/Makefile.mk
+++ b/src/arch/Makefile.mk
@@ -7,6 +7,7 @@ abt_sources += \
 	arch/abtd_affinity.c \
 	arch/abtd_affinity_parser.c \
 	arch/abtd_env.c \
+	arch/abtd_futex.c \
 	arch/abtd_stream.c \
 	arch/abtd_time.c \
 	arch/abtd_ythread.c
--- a/src/arch/abtd_futex.c
+++ b/src/arch/abtd_futex.c
@@ -47,6 +47,24 @@ void ABTD_futex_broadcast(ABTD_futex_multiple *p_futex)
             NULL, 0);
 }
 
+void ABTD_futex_suspend(ABTD_futex_single *p_futex)
+{
+    /* Wake-up signal is 1. */
+    while (ABTD_atomic_acquire_load_int(&p_futex->val) == 0) {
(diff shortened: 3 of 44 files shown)


shintaro-iwasaki:
test:argobots/all
test:argobots/osx
test:argobots/freebsd
test:argobots/solaris

shintaro-iwasaki:
test:argobots/osx

shintaro-iwasaki:
test:argobots/osx

shintaro-iwasaki:
test:argobots/freebsd
test:argobots/solaris
test:argobots/all

shintaro-iwasaki:
We understand that the current synchronization mechanism has room for performance improvement. Let us know any use case so that we can optimize it.

PR #309, opened 2021-03-12
fix memory leak

### LINKED ISSUE #116: malloc failures should be checked and assertion should be avoided

shintaro-iwasaki:
### Problem

Resource allocation errors do not usually occur in testing, so these error handling paths are not tested. As a result, some Argobots functions leak resources, exit while keeping a lock, or cause a double-free corruption when such an error happens.

For example, `ABT_init()` (the default configuration) calls 17 `malloc()` / `mmap()` operations. Usually all the operations succeed (since a machine has enough memory), but Argobots is not tested if any of them fail.  `ABT_init()` should succeed even if some of those operations fail while it "should" return an error after properly freeing already allocated resources if others fail. We want to fix it, but at the same time we should test if that fix is correct.

### What this PR does

This PR fixes those issues the new testing library found. The new testing mechanism overrides `malloc()`, `calloc()`, `mmap()`, `pthread_xxx_init()`, ... functions by using `dlsym()` (actually `dlvsym()`) and simulates all the possible allocation success/failure patterns while tracking allocated resources to detect memory leak. This adds new 15 tests that cover most of the Argobots functions that incur resource allocation (including `ABT_init()` and `ABT_thread_create()`).

See `test/leakcheck/rtrace.h` for details.

### Impact

This PR fixes several bugs, which is great, but user programs do not usually encounter those issues since those issues happen only when resource allocation fails.

This PR closes #116.

## Checklist

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -875,6 +875,67 @@ AC_CHECK_LIB(pthread, pthread_join)
 # check pthread_barrier
 AC_CHECK_FUNCS(pthread_barrier_init)
 
+# check dlvsym
+ABT_RT_CFLAGS=""
+ABT_RT_LDFLAGS=""
+AC_CHECK_LIB(dl, dlvsym)
+if test x"$ac_cv_lib_dl_dlvsym" = x"yes"; then
+    # dl can be used for testing.
+    # check objdump since it is needed
+    if test x"$(which objdump || true)" != x; then
+        # objdump exists.  Dump GLIBC versions.
+        AC_LINK_IFELSE(
+        [AC_LANG_PROGRAM([
+            #include <stdio.h>
+            #include <stdlib.h>
+            #include <sys/mman.h>
+            #include <pthread.h>
+            #include <stdint.h>
+            #include <unistd.h>
+        ],[[
+            typedef void *(*fty)(void *);
+            volatile fty fs[] = {
+                (fty)malloc, (fty)calloc, (fty)realloc, (fty)posix_memalign,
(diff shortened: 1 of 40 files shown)


shintaro-iwasaki:
test:argobots/all
test:argobots/freebsd
test:argobots/solaris
test:argobots/osx

shintaro-iwasaki:
test:argobots/all
test:argobots/freebsd
test:argobots/solaris
test:argobots/osx

shintaro-iwasaki:
test:argobots/freebsd

shintaro-iwasaki:
test:argobots/freebsd

shintaro-iwasaki:
test:argobots/all
test:argobots/solaris
test:argobots/osx
test:argobots/freebsd

shintaro-iwasaki:
test:argobots/all
test:argobots/solaris
test:argobots/osx
test:argobots/freebsd

PR #312, opened 2021-03-14
pool: check the pool's emptiness before taking a lock

shintaro-iwasaki:
## Pull Request Description

`pool_pop()` of the current builtin pool implementations takes a look and then checks the pool size, but this lock is unnecessary if the pool is empty.  This lock acquisition causes significant performance degradation when there are fewer producers than consumers; `pool_push()` operations are delayed by the `pool_pop()`'s lock although that `pool_pop()` will fail since the pool is empty.

This patch adds a quick size check before taking a lock so that `pool_pop()` can return without taking a lock if the pool is empty.

## Performance impact

In the benchmark, only one execution stream creates 10000 ULTs.  All the execution streams access the same pool to execute the created empty ULTs.  The following result shows the relative execution time while the performance of the current main Argobots is 1 (so smaller is better). If more execution streams are created, most pool operations become fruitless `pool_pop()` that returns `ABT_UNIT_NULL`.

image

Briefly speaking, the result shows the following:
1. If there's no contention, there exists a very tiny performance overhead (since this PR adds an extra size check).
2. If there's contention, it can achieve 3 - 25x performance improvement.
3. Using a different cache line for the `is_empty` variable does not improve the performance.  It increases the size of the pool structure, so this alignment is disabled.

This check should be beneficial.

<details><summary>(The benchmark code is collapsed)</summary>
<p>

```c
/* gcc -O3 test.c -labt */

#include <abt.h>
#include <stdio.h>
#include <stdlib.h>

void thread_func(void *arg)
{
    ;
}

int main(int argc, char *argv[])
{
    /* Initialize */
    if (argc != 5) {
        printf("Usage: benchmark.out NUM_XSTREAMS NUM_THREADS NUM_REPEATS "
               "NUM_WARMUPS\n");
        return -1;
    }
    int num_xstreams = atoi(argv[1]);
    int num_threads = atoi(argv[2]);
    int num_repeats = atoi(argv[3]);
    int num_warmups = atoi(argv[4]);

    ABT_init(argc, argv);

    /* Create execution streams. */
    ABT_pool pool;
    ABT_xstream *xstreams =
        (ABT_xstream *)malloc(num_xstreams * sizeof(ABT_xstream));
    ABT_xstream_self(&xstreams[0]);
    ABT_xstream_get_main_pools(xstreams[0], 1, &pool);
    for (int i = 1; i < num_xstreams; i++) {
        /* All execution streams access the same pool. */
        ABT_sched sched;
        ABT_sched_create_basic(ABT_SCHED_DEFAULT, 1, &pool,
                               ABT_SCHED_CONFIG_NULL, &sched);
        ABT_xstream_create(sched, &xstreams[i]);
    }

    ABT_thread *threads =
        (ABT_thread *)malloc(num_threads * sizeof(ABT_thread));

    double start_time = 0.0;
    for (int step = 0; step < num_warmups + num_repeats; step++) {
        if (step == num_warmups)
            start_time = ABT_get_wtime();
        /* Fork threads */
        for (int i = 0; i < num_threads; i++)
            ABT_thread_create(pool, thread_func, NULL, ABT_THREAD_ATTR_NULL,
                              &threads[i]);
        /* Join threads */
        for (int i = 0; i < num_threads; i++)
            ABT_thread_free(&threads[i]);
    }
    double end_time = ABT_get_wtime();

    free(threads);
    printf("Per thread: %.2f [ns]\n",
           (end_time - start_time) / (num_threads * num_repeats) * 1.0e9);

    /* Join ESs */
    for (int i = 1; i < num_xstreams; i++) {
        while (1) {
            /* Only the primary ES can call ABT_finalize() */
            ABT_bool is_primary;
            ABT_self_on_primary_xstream(&is_primary);
            if (!is_primary) {
                ABT_self_yield();
            } else {
                break;
            }
        }
        ABT_xstream_free(&xstreams[i]);
    }
    free(xstreams);
    ABT_finalize();

    return 0;
}
```

<p>
</details>

<details><summary>(The experimental setting is collapsed)</summary>
<p>

Machines:
- Summit-like POWER 9 (IBM S822LC 10 cores / 80 hardware threads * 2 sockets) / RedHat 7.6 / GCC 9.3
- 64-bit ARM (JLSE's 64-bit ARM for HPC, in total 112 hardware threads)/ RedHat 7.6 / GCC 9.3
- Intel Skylake (Intel Xeon 8180M, 28 cores / 56 hardware threads * 2 sockets) / openSUSE 42.2 / GCC 9.3

`./benchmark.out NUM_ESS 10000 1000 100` where NUM_ESS is changed.  Argobots was compiled with `--enable-perf-opt`.  `(this PR + alignment)` means that `is_empty` is stored in another cache line of pool's `data_t` so that accessing `is_empty` does not interfere `pool_push()'s` update.

<p>
</details>

Note that in any case, the absolute performance is terrible if such contention happens.  The user should avoid such a case.

## Checklist

### DIFF
--- a/src/pool/fifo.c
+++ b/src/pool/fifo.c
@@ -32,6 +32,9 @@ struct data {
     size_t num_threads;
     ABTI_thread *p_head;
     ABTI_thread *p_tail;
+    /* If the pool is empty, pop() accesses only is_empty so that pop() does not
+     * slow down a push operation. */
+    ABTD_atomic_int is_empty; /* Whether the pool is empty or not. */
 };
 typedef struct data data_t;
 
@@ -105,6 +108,7 @@ static int pool_init(ABT_pool pool, ABT_pool_config config)
     p_data->num_threads = 0;
     p_data->p_head = NULL;
     p_data->p_tail = NULL;
+    ABTD_atomic_relaxed_store_int(&p_data->is_empty, 1);
 
     p_pool->data = p_data;
 
@@ -141,6 +145,8 @@ static void pool_push_shared(ABT_pool pool, ABT_unit unit)
         p_thread->p_next = p_thread;
         p_data->p_head = p_thread;
         p_data->p_tail = p_thread;
+        p_data->num_threads = 1;
+        ABTD_atomic_release_store_int(&p_data->is_empty, 0);
     } else {
         ABTI_thread *p_head = p_data->p_head;
         ABTI_thread *p_tail = p_data->p_tail;
@@ -149,8 +155,8 @@ static void pool_push_shared(ABT_pool pool, ABT_unit unit)
         p_thread->p_prev = p_tail;
         p_thread->p_next = p_head;
         p_data->p_tail = p_thread;
+        p_data->num_threads++;
     }
-    p_data->num_threads++;
 
     ABTD_atomic_release_store_int(&p_thread->is_in_pool, 1);
     ABTD_spinlock_release(&p_data->mutex);
@@ -167,6 +173,8 @@ static void pool_push_private(ABT_pool pool, ABT_unit unit)
         p_thread->p_next = p_thread;
         p_data->p_head = p_thread;
         p_data->p_tail = p_thread;
+        p_data->num_threads = 1;
+        ABTD_atomic_release_store_int(&p_data->is_empty, 0);
     } else {
         ABTI_thread *p_head = p_data->p_head;
         ABTI_thread *p_tail = p_data->p_tail;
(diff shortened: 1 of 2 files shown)


shintaro-iwasaki:
test:argobots/all
test:argobots/solaris
test:argobots/osx
test:argobots/freebsd

shintaro-iwasaki:
Adding the emptiness check in the spinlock loop  (see the second commit) slightly improves the performance when the lock is contended (yellow (with the second commit) vs orange (without the second commit))
image

shintaro-iwasaki:
test:argobots/all
test:argobots/solaris
test:argobots/osx
test:argobots/freebsd

PR #317, opened 2021-03-24
fix unit-pool mapping

### LINKED ISSUE #316: unit: incorrect unit-pool association management
*This happens only when the user uses a user-defined pool (i.e., `ABT_pool_create()`, not `ABT_pool_create_basic()`).*

## Problem

Whenever Argobots changes the unit-pool association (e.g., by using `ABT_unit_set_associated_pool()` or `ABT_xstream_set_main_sched()`), the unit associated with the old pool must be freed (`u_free()`) and its thread's "unit" must be renewed by a new pool (`u_create_from_thread()`).  However, Argobots does not do this, and it's not something that we can fix by calling `u_free()`.  This is also an API-level issue: `ABT_unit_set_associated_pool(ABT_unit unit, ABT_pool pool)` changes the associated pool of \c unit to \c pool, but then \c unit can be changed according to the new pool's `ABT_unit` representation.

There's no problem if the user uses built-in pools since built-in pools share the data structure of `ABT_unit` and do not allocate memory for each `ABT_thread`.  It becomes problematic when the user uses a user-defined pool.

Currently, not many users use a user-defined pool (e.g., `ABT_pool_create()`), so I do not want to change the major Argobots API or degrade the performance of Argobots for a user-defined pool.  At the same time, this bug inhibits any use of a user-defined pool, so it should be fixed before Argobots 1.1.  It should be fixed soon.

#### What is `ABT_unit`

`ABT_unit` is a scheduling unit that each `ABT_pool` defines.  `ABT_unit` encapsulates either `ABT_thread` or `ABT_task`.  When it was designed, `ABT_unit` can encapsulate more (e.g., `ABT_event` (which does not exist now) or `ABT_sched`), but the Argobots development is towards unifying them; everything becomes `ABT_thread`.  Nonetheless, `ABT_unit` is not the same as `ABT_thread`.

shintaro-iwasaki:
This patch fixes #316.

## Problem

Whenever Argobots changes the unit-pool association (e.g., by using `ABT_unit_set_associated_pool()`, `ABT_xstream_set_main_sched()`, `ABT_pool_push()`, or `ABT_xstream_run_unit()`), the unit associated with the old pool must be freed (`u_free()`) and its thread's "unit" must be renewed by a new pool (`u_create_from_thread()`).

Argobots does not do this, which leaks resources and/or calls pool/unit functions that are not associated with a given `ABT_unit`.  This is also an API-level issue: `ABT_unit_set_associated_pool(ABT_unit unit, ABT_pool pool)` changes the associated pool of \c unit to \c pool, but then \c unit should be changed according to the new pool's `ABT_unit` representation.  Since this function does not take an old pool, the Argobots runtime cannot find a proper `u_free()` to free this the given `unit`.

This is not observed if the user uses a built-in pool (i.e., a pool that is *not* created by `ABT_pool_create()`).

## Solution

This PR fixes this issue by a new `ABT_unit` - `ABTI_pool` (via `ABTI_thread`) hash table so that the runtime can get the pool information from `ABT_unit`. This PR adds a new restriction to `ABT_unit` that `u_create_from_thread()` can return.
1. `ABT_unit` must be 4-byte aligned (so that Argobots can internally use the first 2 bits for built-in pools).
2. The value of `ABT_unit` is unique across systems.  For example, it can be `ABT_thread` or any memory region allocated by `malloc()`.

Specifically, Argobots internally has a hash table for `ABT_unit` and user-defined `ABTI_pool` mapping.  To use a hash table, `ABT_unit` must be unique.  `ABTI_unit` must be 4-byte aligned (or the first two bits are zero) so that the built-in `ABT_unit`, which most Argobots programs use, can be quickly distinguished and avoid using a relatively heavy hash table.

`ABT_unit_set_associated_pool()` becomes an empty function, but it does not change the existing program.  A typical use case of `ABT_unit_set_associated_pool()` should be as follows:
```c
ABT_unit unit;
ABT_pool_pop(remote_pool, &unit)
if (unit != ABT_UNIT_NULL) {
    ABT_unit_set_associated_pool(unit, local_pool);
    ABT_xstream_run_unit(unit, local_pool);
}
```
In this case, `ABT_xstream_run_unit()` will change the `unit`'s pool, so `ABT_unit_set_associated_pool()` gets unnecessary.
```c
ABT_unit unit;
ABT_pool_pop(remote_pool, &unit)
if (unit != ABT_UNIT_NULL) {
    // ABT_unit_set_associated_pool is not harmful, though.
    // ABT_unit_set_associated_pool(unit, local_pool);
    ABT_xstream_run_unit(unit, local_pool);
}
```

This automatic unit-pool association is done for all functions that change the unit/thread-pool mapping such as `ABT_xstream_run_unit()`, `ABT_pool_push()`, `ABT_thread_set_associated_pool()`, `ABT_xstream_set_main_sched()`, and `ABT_thread_revive()`.

### Performance Impact

This adds several branches, so this change should negatively affect the performance even for built-in pools.  To compensate the performance degradation, this PR inlines some unit functions, which would improve the performance.  Overall, the performance difference should be little.  In any case, this is for correctness and no matter how much it affects the performance, this fix should be merged.

If the user uses a user-defined pool, this PR adds a quite large overhead, but without this PR, a user-defined pool is not functional (unlike built-in pools, which work well without this PR).

### DIFF
--- a/Doxyfile.in
+++ b/Doxyfile.in
@@ -1902,6 +1902,8 @@ ALIASES += DOC_DESC_V10_CRUDE_TERMINATION_CHECK{2}="\DOC_V10 \2 is returned if t
 
 ALIASES += DOC_DESC_V10_REJECT_MUTEX_ATTR_NULL{1}="\DOC_V10 This routine returns \c ABT_ERR_INV_MUTEX_ATTR if \1 is \c ABT_MUTEX_ATTR_NULL.\n \DOC_V11 This routine uses the default mutex attributes if \1 is \c ABT_MUTEX_ATTR_NULL. @rationale Most of the Argobots routines use the default attributes when a NULL handle is passed as a configuration or an attribute.  This routine should follow this behavior. @endrationale"
 
+ALIASES += DOC_DESC_V10_UNIT_POOL_MAPPING="\DOC_V10 The user must explicitly call this routine to change the associated pool of \c ABT_unit.\n \DOC_V11 The Argobots runtime automatically changes the associated pool of \c ABT_unit if needed.  This routine is therefore unnecessary. @rationale The unit-pool association management is hard for the user who does not know the internal implementation of Argobots.  This change not only eases programming with Argobots but also addresses some issues without breaking the backward compatibility.  See https://github.com/pmodels/argobots/issues/317 for details. @endrationale"
+
 ALIASES += DOC_DESC_V1X_ERROR_CODE_CHANGE{3}="\DOC_V1X \1 is returned if \3.\n \DOC_V20 \2 is returned if \3. @rationale Argobots 2.0 changes the error code for consistent behavior with other Argobots routines. @endrationale"
 
 ALIASES += DOC_DESC_V1X_NOEXT{1}="\DOC_V1X If an external thread calls this routine, \1 is returned.\n \DOC_V20 An external thread may call this routine. @rationale Argobots 2.0 allows an external thread where possible unless it is semantically inappropriate. @endrationale"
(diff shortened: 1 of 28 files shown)


shintaro-iwasaki:
test:argobots/all
test:argobots/solaris
test:argobots/osx
test:argobots/freebsd

shintaro-iwasaki:
test:argobots/all

shintaro-iwasaki:
test:argobots/all
test:argobots/solaris
test:argobots/osx
test:argobots/freebsd

shintaro-iwasaki:
test:argobots/all
test:argobots/solaris
test:argobots/osx
test:argobots/freebsd

shintaro-iwasaki:
It looks working after fixing #318. I'll merge this PR.

PR #318, opened 2021-03-24
xstream: fix xstream_set_main_sched

shintaro-iwasaki:
## Pull Request Description

This bug happens when `ABT_xstream_set_main_sched()` is called to update the underlying execution stream's main scheduler and the new scheduler's pool is shared with other schedulers associated with already running execution streams.

In this case, `ABT_xstream_set_main_sched()` pushes the current ULT to the pool **before** context switch, so this current ULT might be executed by other execution streams before saving its context by `ABT_xstream_set_main_sched()`, crashing a program. 

This patch fixes it by pushing this ULT to the pool **after** context switch.

## Checklist


shintaro-iwasaki:
test:argobots/all
test:argobots/solaris
test:argobots/osx
test:argobots/freebsd

Issue #324, opened 2021-04-04
How to detect stack overflow


xuqing-er:
Hi~
  I found that in libgo, libcopp and other coroutine libraries, guard pages are used to detect stack overflow. I want to know whether argobots's detection mechanism is the same as them?

thanks

shintaro-iwasaki:
Currently Argobots supports only stack canary (i.e., no active mechanism that uses `mprotect()` or something similar).

It's discussed in #274. Please check it if you are interested in this issue. I'd be happy to hear any suggestions.

PR #327, opened 2021-04-19
thread: support mprotect-based stack guard

shintaro-iwasaki:
## Pull Request Description

As requested by many users (see #274, #324, or [this thread](https://lists.argobots.org/pipermail/discuss/2021-April/000113.html)), this patch introduces an `mprotect()`-based stack guard.  The program will receive SEGV if the program accesses the protected 4KB page located at the bottom of the ULT stack.

#### Method A.
Pass `--enable-stack-overflow-check=mprotect` at configure time.  This enables the `mprotect()`-based stack guard by default.
```sh
./configure --enable-stack-overflow-check=mprotect
```

***OR***

#### Method B.
Set `ABT_STACK_OVERFLOW_CHECK=mprotect` at execution time.  This **does not** require a configure-time setting.
```sh
ABT_STACK_OVERFLOW_CHECK=mprotect ./prog.out
```

### Expected behavior.

When the program accesses the protected 4KB page located at the bottom of the ULT stack, the program will receive SEGV.  The user cannot continue this program (since it accesses an invalid memory region), but the user can know which thread crushes the stack.  The user should increase the stack size (e.g., setting `ABT_THREAD_STACKSIZE=XXX` for ULTs with the default attribute) and run this program again.

### Resource consumption by `mprotect()`.

In a typical environment this `mprotect()` works only for the 30,000 ULTs.

One distinct `mprotect()` region consumes an `mmap()` system resource, while typically the system allows the process to create at most 60,000 regions (see `sysctl vm.max_map_count`).  Since this `mprotect()` mechanism creates protected and unprotected regions per ULT, the user can use this mechanism for the first 30,000 regions.

By default, Argobots ignores the failure of `mprotect()`, which happens when the user maintains more than 30,000 ULTs, but at this point, the user should be aware that the program cannot allocate new `mmap()`'ed regions, so an error can happen in different places.

Alternatively, the user can set `ABT_STACK_OVERFLOW_CHECK=mprotect_strict` (or pass `--enable-stack-overflow-check=mprotect-strict`) to assert this `mprotect()` error.  Currently, this error is unrecoverable; the Argobots runtime aborts.

Most programs do not create 30,000 ULTs, but if the program manages more than 30,000 ULTs at the same time (precisely speaking, `vm.max_map_count / 2` ULTs), the user should consider either disabling this feature or increasing `vm.max_map_count`.

### Stack size

This mechanism uses at most 8KBs (precisely speaking, `system page size * 2`) from the stack.  The default stack size is automatically increased, while the stack size is not increased if the user gives a stack size or a stack memory.  This is the user's responsibility to set the proper stack size.

### Stack canary vs. mprotect-based stack guard.

Stack canary is good because it is lightweight and does not use any system resources.  Stack canary cannot pinpoint when the ULT crashes the stack, though.  Sometimes another problem is caused before the runtime detects the death of the stack canary.  This stack canary is useful to check if there is no problem: if a stack canary is alive, the user can know that the program did not break the ULT stacks.  To use stack canary, set `--enable-stack-overflow-check=canary` or `--enable-stack-overflow-check=canary-XXX` where `XXX` is the canary size.

`mprotect()`-based stack guard is good for debugging since the user can know where the stack smash happens.  However, this is heavy (especially when the ULT stack is not obtained from a built-in stack pool) and limited by `mmap` resources.  This should be useful for debugging.

### Performance impact.

If `mprotect()`'ed stacks are cached.  There's no overhead.  A pair of `mprotect()` costs around 1 microsecond on an enterprise Intel machine, so an additional overhead is 1 microsecond per ULT fork-join if that ULT is not cached (e.g., a user-given stack size / stack memory).  The Argobots ULT fork-join operation can be a few hundred nanoseconds (except for external `malloc()`/`free()` overheads), this microsecond-order cost is not very small.  The user should be aware of this overhead.

## Checklist


shintaro-iwasaki:
test:argobots/all
test:argobots/freebsd
test:argobots/osx
test:argobots/solaris

shintaro-iwasaki:
test:argobots/all
test:argobots/freebsd
test:argobots/osx
test:argobots/solaris

shintaro-iwasaki:
test:argobots/all
test:argobots/freebsd
test:argobots/osx
test:argobots/solaris

PR #328, opened 2021-04-20
thread: fix mprotect-based stack guard 

shintaro-iwasaki:
## Pull Request Description

PR #327 did not consider a system that has larger page sizes. For example, typically Intel x86/64 machines use 4KB pages while the other systems such as 64-bit ARM and POWER9 sometimes use 64KB pages or larger. This PR fixes the test as well as the stack size adjustment to support those systems.

This PR does not affect the behavior of Argobots unless you enable this stack guard.

## Checklist


shintaro-iwasaki:
test:argobots/all
test:argobots/freebsd
test:argobots/osx
test:argobots/solaris

PR #331, opened 2021-04-23
util: fix an argument of mmap

shintaro-iwasaki:
## Pull Request Description

`fd` must be `-1` when `MAP_ANONYMOUS` or `MAP_ANON` is set for portability. For example, `mmap()` fails on FreeBSD if `fd` is `0`.  Note that Linux accepts both `0` and `-1`.

Linux `mmap()`:
https://man7.org/linux/man-pages/man2/mmap.2.html
FreeBSD `mmap()`:
https://www.freebsd.org/cgi/man.cgi?query=mmap&sektion=2&format=html

## Checklist

### DIFF
--- a/src/util/largepage.c
+++ b/src/util/largepage.c
@@ -33,7 +33,7 @@
 static void *mmap_regular(size_t size)
 {
 #if ABTU_LP_USE_MMAP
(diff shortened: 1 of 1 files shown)


shintaro-iwasaki:
test:argobots/all
test:argobots/freebsd
test:argobots/osx
test:argobots/solaris

PR #334, opened 2021-06-02
waitlist: fix ABTI_waitlist_is_empty

shintaro-iwasaki:
## Pull Request Description

`ABTI_waitlist_is_empty()` returned `ABT_TRUE` if the `ABTI_waitlist` is "not" empty, but it worked correctly since its caller (`ABT_cond_free()`) also misused this function as if this function is "`ABTI_waitlist_not_empty()`". 

This patch fixes the implementation of `ABTI_waitlist_is_empty()` so that its name correctly explains what its body.

Note that this fix does not affect the behavior of Argobots.

## Checklist


shintaro-iwasaki:
test:argobots/all

PR #335, opened 2021-06-02
error: assert when undefined behavior may occur

### LINKED ISSUE #315: assert if possible when undefined behavior occurs
Argobots defines when undefined behaviors can happen, but this is not very user friendly. If possible, Argobots should warn users when a debug option is true so that user can easily find and fix the cause of undefined behavior. The idea is that, since the behavior is undefined in such a case, that behavior can be an assertion when a debug option is enabled. Undefined-behavior checks are expensive, so they will be removed for a non-debug configuration.

Examples of when undefined behaviors happen:
- Passing `NULL`
- Double-freeing an Argobots object (e.g., `ABT_pool`)
- Double-joining a ULT
- Explicitly freeing an automatic scheduler/pool
- Explicitly freeing a pool of the primary execution stream.

Note that this detection is neither mandatory nor strict since it causes "undefined behavior". Argobots should do the best to show a useful error message rather than incur an error somewhere else, but the user should not expect assertion.

See this thread for related discussion: https://lists.argobots.org/pipermail/discuss/2021-March/000103.html

shintaro-iwasaki:
## Pull Request Description

This PR closes #315.

Argobots defines when undefined behaviors may happen but the runtime does not actively detect such since such a check is either mostly unnecessary (e.g., check if Argobots is initialized or not) or extremely costly (e.g., detection of double-freeing an Argobots object). This PR provides an option to check some of them on a best efforts basis.

Even after this PR, Argobots checks only runtime initialization and NULL arguments in most places. Some checks regarding synchronization objects, however, would be generally useful (e.g., check if there's no waiter when freeing `ABT_eventual`). We might add more expensive checks if needed.

Note that this does not break the current API since "undefined behavior" can cause anything, including failure of `assert()`. At the same time, please do not rely on this assertion behavior since at any rate this is part of undefined behavior. These checks can be removed or relaxed in the future for code readability or performance optimization.

## Checklist

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -95,14 +95,15 @@ AC_ARG_ENABLE([affinity],
 AC_ARG_ENABLE([debug],
 [  --enable-debug@<:@=OPTS@:>@   control the level of debugging. "OPTS" is a list of
                           comma separated names below. Default is "yes".
-        yes   - add compiler flags, -g -Wall
-        err   - print abt_errno information on standard error
-        log   - print debug event logging
-                (set ABT_USE_LOG to 0 to disable logging)
-        all   - all of the above choices (equal to "yes,err,log")
-        most  - same as "all" but not print logs by default
-                (set ABT_USE_LOG to 1 to enable logging)
-        none  - no debugging, i.e., --disable-debug
+        yes       - add compiler flags, -g -Wall
+        err       - print abt_errno information on standard error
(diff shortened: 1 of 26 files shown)


shintaro-iwasaki:
test:argobots/all
test:argobots/osx
test:argobots/solaris
test:argobots/freebsd

PR #336, opened 2021-06-16
cleanup fcontext support

shintaro-iwasaki:
## Pull Request Description

This PR updates fcontext support to examine and unify the interface of the current fcontext implementations. Specifically, this patch removes the following fcontext implementations that were not well tested:

#### Darwin (OSX, Mac):
- i386 (32-bit Intel)
- ppc32 (32-bit POWER)
- ppc64 (64-bit POWER)

#### ELF (Linux, FreeBSD)
- ppc32 (32-bit POWER)

I do not think no one uses Argobots on one of the platforms I wrote above (note, Mac+64-bit Intel will be kept supported). Note that Argobots continues to support ucontext, so Argobots should be functional on those platforms even without fcontext support.

This PR also fixes configure.ac to support FreeBSD + 64-bit POWER.

## Checklist


shintaro-iwasaki:
test:argobots/all
test:argobots/freebsd
test:argobots/solaris
test:argobots/osx

PR #339, opened 2021-06-28
log: more consistent logging

shintaro-iwasaki:
## Pull Request Description

Argobots has a logging feature, but `LOG_DEBUG()` is not maintained well. For example, it sometimes misses a counterpart of operations (e.g., create/terminate or suspend/resume). The problem is that we do not test log messages, so its maintenance is done in an ad-hoc manner.

This patch uses `ABTI_event` (previously `ABTI_tool_event`) to print logs. This logging feature is more reliable since the timing of event invocation is tested via is tested via the tool interface.

This patch does not affect the behavior of Argobots if the user does not turn on the logging feature.

## Checklist


shintaro-iwasaki:
test:argobots/all

shintaro-iwasaki:
test:argobots/all

shintaro-iwasaki:
test:argobots/all

Issue #155, opened 2020-03-17
Work-first policy at thread creation


s417-lama:
Currently Argobots supports only help-first policy (push a new thread to a queue and run continuation), but sometimes work-first policy (run a new thread and push the current thread to a queue) is useful for recursive parallelism with work stealing.

Even though the queue interface (#154) is improved, work-first is hard to realize by combining `ABT_thread_create` and `ABT_thread_yield`.
 I want something like `ABT_thread_create_ex` with an argument to tell if it is work-first or not.

shintaro-iwasaki:
I understand this need; it has been also asked by several people. 

The Argobots team experimentally implemented work-first (i.e., child-first) ULTs in https://ieeexplore.ieee.org/document/9018074/. The performance reported in this article looks good; work-first thread creation is generally good for preserving locality and in particular it is used for deep and narrow parallelism (such as N-way recursive divide-and-conquer parallelism where N is not large).

The interface should be discussed more (e.g., `ABT_thread_create_ex` or new `ABT_thread_attr` or others) but in my opinion supporting work-first is beneficial.

PR #342, opened 2021-07-06
improve ULT request handling mechanism

shintaro-iwasaki:
## Problem

Some threading functions need to perform an operation to complete *after context switching*.  For example, a ULT that yields needs to be pushed back to its associated pool after it switches to its parent ULT (otherwise, another scheduler accesses that pool and executes that ULT which is still running, crashing the program).  Such an operation across context switch is achieved by request handling: before a ULT yields, this ULT sets a "POOL_PUSH" request, and the resuming ULT (= the parent scheduler) executes a pool operation based on its request.

This request handling mechanism is very complicated since the developer does not know which request handling would be performed by which ULT.  For example, a join-termination request attached to a joinee (i.e., a ULT that finishes) is handled only by a joiner (i.e., a ULT that calls `ABTI_thread_join()`), but this flow is difficult to follow since those are implemented in different files.  The algorithm is often complicated since some request handling operations include atomic operations (e.g., setting a ULT that suspends to resumable after context switch).  This is similar to a critical section that starts and finishes in different functions (that are running on different ULTs). making the logic extremely hard to understand.  It increases the maintenance cost and prevents flexibility.

## Solution

This PR introduces a callback mechanism to perform request handling.  The next ULT just runs a callback function without knowing its detail (i.e., without knowledge of requests that are possibly set to the previous ULT).

Passing a callback function to the next ULT is heavy, so we introduce a new fcontext implementation that embeds calling this callback operation.

This PR is not divisible since this PR changes the core algorithm in Argobots including the raw fcontext implementation.
In any case, those technical details do not matter any threading features supported by Argobots.  

### Performance

First of all, this patch is for code maintenance, so performance improvement is not the primary goal.  Potentially this PR can degrade the performance because of function calling overheads (the request handling code was inlined) while it improves since it removes branches (the request handling code has several branches to support various requests).

The following shows the performance.
 - `Main`: the current Argobots
 - `-O2`: the default configuration
 - `opt`: `--enable-perf-opt`

Higher is better.  The results show that the performance is almost the same (in most cases, within 5%), which should be acceptable.

image

<details><summary> Benchmark code (collapsed)</summary>
<p>

```c
#include <abt.h>
#include <stdio.h>
#include <stdlib.h>
#include <assert.h>

typedef struct {
    ABT_thread thread;
    int num_yields;
} thread_arg_t;

void thread_func(void *arg)
{
    thread_arg_t *p_arg = (thread_arg_t *)arg;
    const int num_yields = p_arg->num_yields;
    for (int i = 0; i < num_yields; i++) {
        ABT_thread_yield();
    }
}

{
    if (argc != 3) {
        printf("Usage: ./a.out NUM_THREADS NUM_YIELDS\n");
        return -1;
    }
    const double min_repeat_sec = 1.0; // [s]
    const int num_min_repeats = 10;
    const int num_warmups = 3;
    const int num_threads = atoi(argv[1]);
    const int num_yields = atoi(argv[2]);

    ABT_init(0, NULL);
    thread_arg_t *thread_args =
        (thread_arg_t *)malloc(sizeof(thread_arg_t) * num_threads);
    for (int i = 0; i < num_threads; i++) {
        thread_args[i].num_yields = num_yields;
    }

    // Use a private pool since it is the fastest.
    ABT_pool pool;
    ABT_pool_create_basic(ABT_POOL_FIFO, ABT_POOL_ACCESS_PRIV, ABT_TRUE, &pool);
    ABT_sched sched;
    ABT_sched_create_basic(ABT_SCHED_BASIC, 1, &pool, ABT_SCHED_CONFIG_NULL,
                           &sched);
    ABT_xstream xstream;
    ABT_self_get_xstream(&xstream);
    ABT_xstream_set_main_sched(xstream, sched);
    // Measurement
    double start_time, end_time;
    int repeat = 0, target_repeat = num_warmups + num_min_repeats;
    while (1) {
        repeat++;
        if (repeat == num_warmups)
            start_time = ABT_get_wtime();
        // Kernel starts
        for (int i = 0; i < num_threads; i++) {
            ABT_thread_create(pool, thread_func, &thread_args[i],
                              ABT_THREAD_ATTR_NULL, &thread_args[i].thread);
        }
        for (int i = num_threads - 1; i >= 0; i--) {
            ABT_thread_free(&thread_args[i].thread);
        }
        // Kernel ends
        if (repeat == target_repeat) {
            end_time = ABT_get_wtime();
            if (end_time - start_time > min_repeat_sec) {
                break;
            } else {
                target_repeat *= 2;
            }
        }
    }
    printf("[%d, %d] %f [ns per ULT]\n", num_threads, num_yields,
           (end_time - start_time) / num_threads / (repeat - num_warmups) *
               1.0e9);
    free(thread_args);
    ABT_finalize();
    return 0;
}
```

</p></details>

<details><summary> Experimental setting (collapsed)</summary>
<p>

All experiments used 1 core.
x86/64: Intel Skylake 8180 (openSUSE 15.2, gcc 7.5.0)
ARM: ThunderX2 (RedHat 7.6, gcc 9.3.0)
POWER: Summit-like POWER 9 machine (RedHat, gcc 9.3.0)
Average of 10 executions.
Fork-join benchmark: # of yields = 0
Yield benchmark: # of ULTs = 8
</p></details>

This PR is necessary to implement work-first scheduling (#155).

## Checklist

### DIFF
--- a/src/arch/abtd_ythread.c
+++ b/src/arch/abtd_ythread.c
@@ -30,14 +30,12 @@ static inline void ythread_terminate(ABTI_xstream *p_local_xstream,
     ABTD_ythread_context *p_link =
         ABTD_atomic_acquire_load_ythread_context_ptr(&p_ctx->p_link);
     if (!p_link) {
-        uint32_t req =
-            ABTD_atomic_fetch_or_uint32(&p_ythread->thread.request,
-                                        ABTI_THREAD_REQ_JOIN |
-                                            ABTI_THREAD_REQ_TERMINATE);
+        uint32_t req = ABTD_atomic_fetch_or_uint32(&p_ythread->thread.request,
+                                                   ABTI_THREAD_REQ_JOIN);
         if (!(req & ABTI_THREAD_REQ_JOIN)) {
             /* This case means there is no join request.  Let's go back to the
              * parent ULT */
-            ABTI_ythread_jump_to_parent(p_local_xstream, p_ythread);
+            ABTI_ythread_terminate(p_local_xstream, p_ythread);
             ABTU_unreachable();
         } else {
             /* This case means there has been a join request and the joiner has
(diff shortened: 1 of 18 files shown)


shintaro-iwasaki:
test:argobots/all
test:argobots/solaris
test:argobots/freebsd
test:argobots/osx

PR #343, opened 2021-07-06
self: suppress warning message of ABT_self_get_type

### LINKED ISSUE #341: ABT_self_get_type() printing on type == ABT_UNIT_TYPE_EXT
We have a generic logging callback that tried to log the "thread" identifier so is called in many different contexts and one of the things it does is call ABT_self_get_thread_id(), however that fails if called outside of a argobots context, so before we call that we call ABT_self_get_type() which appears to do exactly what we want.

Looking at the code we're don't have ABT_CONFIG_ENABLE_VER_20_API set, so we're seeing log messages of "[self.c:300] ABT_self_get_type: 4" printed every time we call this function, and it's then returning ABT_ERR_INV_XSTREAM and setting type to ABT_UNIT_TYPE_EXT.

Other than the debug printf when using --enable-debug everything is working as we expect, but it seems that calling this function in this context is OK so the message should not be printed in this case.

One fix would be to create a new ABTI_SETUP_LOCAL_XSTREAM_SILENT() macro to remove the print and call it from self_get_type, there would be the possibility to use this in ABT_self_get_thread_id() as well which would work for our use-case but I can see why you might not want this generally.

shintaro-iwasaki:
## Pull Request Description

This PR fixes #341. Specifically, this PR suppresses a warning message when `ABT_self_get_type()` (Argobots 1.x) is called on an external thread.  This warning message was printed when `--enable-debug=err` (included by `most` or `all`) is set.

(Note that `ABT_initialized()`, `ABT_mutex_trylock()`, and `ABT_cond_timedwait()` also legally return an error like `ABT_self_get_type()`. They do not print a warning message.)

## Checklist

### DIFF
--- a/src/self.c
+++ b/src/self.c
@@ -296,8 +296,14 @@ int ABT_self_get_type(ABT_unit_type *type)
     /* By default, type is ABT_UNIT_TYPE_EXT in Argobots 1.x */
     *type = ABT_UNIT_TYPE_EXT;
     ABTI_SETUP_GLOBAL(NULL);
-    ABTI_xstream *p_local_xstream;
-    ABTI_SETUP_LOCAL_XSTREAM(&p_local_xstream);
+    /* Since ABT_ERR_INV_XSTREAM is a valid return, this should not be handled
+     * by ABTI_SETUP_LOCAL_XSTREAM, which warns the user when --enable-debug=err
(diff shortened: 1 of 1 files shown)


shintaro-iwasaki:
test:argobots/all

PR #344, opened 2021-07-08
self: add ABT_self_yield_to and its friends

shintaro-iwasaki:
## Pull Request Description

This PR implements the following new APIs, which become feasible thanks to #342.

```c
int ABT_self_yield_to(ABT_thread thread);
int ABT_self_suspend_to(ABT_thread thread);
int ABT_self_exit_to(ABT_thread thread);
```

Specifically, `ABT_self_yield_to()` is an important API since it can replace the current `ABT_thread_yield_to()`; `ABT_thread_yield_to()` internally uses `pool_remove()`, which is **not thread-safe**, so the target ULT of `ABT_thread_yield_to()` must be in a pool that is accessed by no other ULTs/execution streams in parallel.  This restriction makes `ABT_thread_yield_to()` useless.

To do something similar to `ABT_thread_yield_to()`, the user can do the following:
```c
// mimic ABT_thread_yield_to()
void thread_yield_to(ABT_thread thread) {
  ABT_pool pool;
  ABT_unit unit;
  ABT_thread_get_last_pool(thread, &pool);
  ABT_thread_get_unit(thread, &unit)
  ABT_pool_remove(pool, unit); // not thread-safe (i.e., if unit is not in the pool, the program crashes)
  ABT_self_yield_to(thread);
}
```

Instead, the user might want to do something like this with `ABT_self_yield_to()`.  The following is an example of priority-based scheduling.
```c
// Directly jump to a ULT in a priority pool (for some reasons)
ABT_unit unit;
ABT_pool_pop(high_prio_pool, &unit); // Thread-safe.
if (unit != ABT_UNIT_NULL) {
  ABT_thread thread;
  ABT_unit_get_thread(unit, &thread);
  // Directly yield to a high-priority ULT.
  ABT_self_yield_to(thread);
}
```

This patch does not affect the existing functions.

## Checklist

### DIFF
--- a/src/arch/abtd_ythread.c
+++ b/src/arch/abtd_ythread.c
@@ -116,6 +116,43 @@ void ABTD_ythread_cancel(ABTI_xstream *p_local_xstream, ABTI_ythread *p_ythread)
     ABTI_event_thread_cancel(p_local_xstream, &p_ythread->thread);
 }
 
+void ABTD_ythread_exit_to(ABTI_xstream *p_local_xstream,
+                          ABTI_ythread *p_cur_ythread,
+                          ABTI_ythread *p_tar_ythread)
+{
+    /* If other ULT is blocked to join the canceled ULT, we have to wake up the
+     * joiner ULT.  However, unlike the case when the ULT has finished its
+     * execution and calls ythread_terminate/exit, this caller of this function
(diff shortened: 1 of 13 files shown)


shintaro-iwasaki:
test:argobots/all
test:argobots/freebsd
test:argobots/solaris
test:argobots/osx

PR #348, opened 2021-07-14
self: add ABT_self_resume_yield_to and its friends

shintaro-iwasaki:
## Pull Request Description

This PR extends #344 by implementing the following new APIs.

```c
int ABT_self_resume_yield_to(ABT_thread thread);
int ABT_self_resume_suspend_to(ABT_thread thread);
int ABT_self_resume_exit_to(ABT_thread thread);
```
The calling ULT resumes a suspended ULT and jumps to it. These routines allow more precise scheduling since the algorithm can describe which ULT should stop (by suspension) and exactly which ULT should be scheduled next. 

They cannot be achieved by combining existing APIs.
```c
// ABT_self_resume_yield_to() does something similar to the following "atomically".
void self_resume_yield_to(ABT_thread thread) {
  ABT_pool pool;
  ABT_unit unit;
  ABT_thread_resume(thread);
  // It looks okay, but ABT_thread_resume() pushes "thread" back to its
  // associated pool, so this "thread" might have already been popped by
  // another scheduler.
  ABT_thread_get_last_pool(thread, &pool);
  ABT_thread_get_unit(thread, &unit)
  ABT_pool_remove(pool, unit);
  ABT_self_yield_to(thread);
}
```

This patch does not affect the existing functions.

## Checklist


shintaro-iwasaki:
test:argobots/all
test:argobots/osx
test:argobots/solaris
test:argobots/freebsd

PR #350, opened 2021-07-14
self: add ABT_self_schedule

shintaro-iwasaki:
## Pull Request Description

This PR implements a new API `ABT_self_schedule()`, which is similar to `ABT_xstream_run_unit()`.
```c
int ABT_self_schedule(ABT_thread thread, ABT_pool pool);
```
The difference is the following two points.
1. `ABT_self_schedule()` takes `ABT_thread`, not `ABT_unit`.
2. `ABT_self_schedule()` accepts `ABT_POOL_NULL`, and if so, pool association of `thread` is not changed.

This PR completes all the basic context switching functions that bypass a scheduler. The new set of API should give users full flexibility of ULT scheduling; it should be especially useful to implement special ULT synchronization mechanisms on top of Argobots.
image

This patch does not affect the existing API.

## Checklist


shintaro-iwasaki:
test:argobots/all
test:argobots/osx
test:argobots/freebsd
test:argobots/solaris

PR #351, opened 2021-07-15
thread: fix asynchronous request handling

shintaro-iwasaki:
## Pull Request Description

Currently, Argobots supports two types of asynchronous requests for threads: cancellation and migration.  Now both are checked on context switching, but those requests can break the algorithm (e.g., cancelling a ULT that is waiting for a synchronization object). This can be the user's responsibility, but the runtime should ignore those requests if it can potentially break the algorithm since migration and cancellation timings are not defined.

This patch fixes this request handling algorithm and cleans it up to make its logic simpler.

As a result, this PR reduces the number of request handling sites, so this PR should not degrade the performance of the Argobots runtime.  This PR should not affect most user programs because most users do not use these features.

If you find any issues, please let us know!

## Checklist

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -179,9 +179,8 @@ AC_ARG_ENABLE([feature],
 [  --enable-feature@<:@=OPTS@:>@  enable/disable features.
                           "OPTS" is a list of comma separated names including
         all                 - all features always enabled (default)
-        no-thread-cancel    - disable ULT cancellation
-        no-task-cancel      - disable tasklet cancellation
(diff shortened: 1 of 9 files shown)


shintaro-iwasaki:
test:argobots/all
test:argobots/osx
test:argobots/freebsd
test:argobots/solaris

shintaro-iwasaki:
test:argobots/osx
test:argobots/freebsd
test:argobots/solaris

PR #356, opened 2021-08-03
thread: lazy stack allocation

shintaro-iwasaki:
## Pull Request Description

This PR implements lazy-stack-allocation (LSA) and eager-stack-release (ESR) optimization in the following paper: https://ieeexplore.ieee.org/document/9018074.

This optimization is disabled by default.  Please turn it on if you want.
```sh
./configure --enable-lazy-stack-alloc
```

The idea of this optimization is assigning ULT stacks to **only active ULTs**.  This method has two merits.
1. Because ULT stacks are associated with ULTs only when needed, it can reduce the ULT stack consumption (or the user can make each stack larger).
2. Because the same ULT stack can be used for consecutively executed ULTs, it can improve the cache locality.

### Idea

Consider the following code:
```c
for (int i = 0; i < 4; i++) {
    ABT_thread_create(..., &threads[i]);
}
for (int i = 0; i < 4; i++) {
    ABT_thread_free(&threads[i]);
}
```

The current Argobots allocates stacks in `ABT_thread_create()`, so 4 ULT stacks are consumed to `threads[0:4]` in the first loop.  Those stacks are freed by `ABT_thread_free()`.

Ultimately, a ULT stack is to keep the intermediate state of a ULT function.  We need to save a state (function call stacks) in a ULT stack if a ULT context switches to another.  However, if those `threads` do not "yield" --just running ULT functions without yield--, we do not need to save such a state.  In this case, as a result, allocating 4 ULT stacks for `threads` are wasteful.

image

If we assign a stack on "scheduling" (`ABT_self_schedue()`) and frees it when a ULT finishes, the code above consumes only 1 ULT stack if a ULT function does not yield.  This does not prevent a ULT from yielding.  The maximum stack consumption does not exceed the stack consumption of the current Argobots (i.e., at most 1 stack per ULT).

### Evaluation

Performance improvement is sometimes significant.

 - `Main`: the current Argobots
 - `-O2`: the default configuration
 - `opt`: `--enable-perf-opt`
 - `w/ Lazy`: this PR + `--enable-lazy-stack-alloc`
 - `wo Lazy`: this PR without `--enable-lazy-stack-alloc`

image

#### Performance
Typically this optimization is beneficial when fewer ULTs yield.  The left figures show a non-yield case; in this case, the ULT overhead gets constant thanks to a better cache locality. The right figures show a yield case; in this case, there is not performance benefit. This PR adds a few branches, so it degrades the performance when a ULT yields.  Note that the performance degradation of yield is negligible in most real applications since the cause of "yield" is often the real performance bottleneck (e.g., mutex, waiting for request completion, frequent ULT synchronization, and so on).

#### Memory footprint
The memory footprint is significantly low; with `--enable-lazy-stack-alloc`, this benchmark uses only 1 ULT stack (16KB if ULT stack is 16KB) for ULTs while it allocates 2 million ULT stacks (32GB in total) without this setting.

Please check the paper above if you are interested in it.

<details><summary> Benchmark code (collapsed)</summary>
<p>

```c
#include <abt.h>
#include <stdio.h>
#include <stdlib.h>
#include <assert.h>

typedef struct {
    ABT_thread thread;
    int num_yields;
} thread_arg_t;

void thread_func(void *arg)
{
    thread_arg_t *p_arg = (thread_arg_t *)arg;
    const int num_yields = p_arg->num_yields;
    for (int i = 0; i < num_yields; i++) {
        ABT_thread_yield();
    }
}

{
    if (argc != 3) {
        printf("Usage: ./a.out NUM_THREADS NUM_YIELDS\n");
        return -1;
    }
    const double min_repeat_sec = 1.0; // [s]
    const int num_min_repeats = 10;
    const int num_warmups = 3;
    const int num_threads = atoi(argv[1]);
    const int num_yields = atoi(argv[2]);

    ABT_init(0, NULL);
    thread_arg_t *thread_args =
        (thread_arg_t *)malloc(sizeof(thread_arg_t) * num_threads);
    for (int i = 0; i < num_threads; i++) {
        thread_args[i].num_yields = num_yields;
    }

    // Use a private pool since it is the fastest.
    ABT_pool pool;
    ABT_pool_create_basic(ABT_POOL_FIFO, ABT_POOL_ACCESS_PRIV, ABT_TRUE, &pool);
    ABT_sched sched;
    ABT_sched_create_basic(ABT_SCHED_BASIC, 1, &pool, ABT_SCHED_CONFIG_NULL,
                           &sched);
    ABT_xstream xstream;
    ABT_self_get_xstream(&xstream);
    ABT_xstream_set_main_sched(xstream, sched);
    // Measurement
    double start_time, end_time;
    int repeat = 0, target_repeat = num_warmups + num_min_repeats;
    while (1) {
        repeat++;
        if (repeat == num_warmups)
            start_time = ABT_get_wtime();
        // Kernel starts
        for (int i = 0; i < num_threads; i++) {
            ABT_thread_create(pool, thread_func, &thread_args[i],
                              ABT_THREAD_ATTR_NULL, &thread_args[i].thread);
        }
        for (int i = num_threads - 1; i >= 0; i--) {
            ABT_thread_free(&thread_args[i].thread);
        }
        // Kernel ends
        if (repeat == target_repeat) {
            end_time = ABT_get_wtime();
            if (end_time - start_time > min_repeat_sec) {
                break;
            } else {
                target_repeat *= 2;
            }
        }
    }
    printf("[%d, %d] %f [ns per ULT]\n", num_threads, num_yields,
           (end_time - start_time) / num_threads / (repeat - num_warmups) *
               1.0e9);
    free(thread_args);
    ABT_finalize();
    return 0;
}
```

</p></details>

<details><summary> Experimental setting (collapsed)</summary>
<p>

All experiments used 1 core.
x86/64: Intel Skylake 8180 (openSUSE 15.2, gcc 7.5.0)
ARM: ThunderX2 (RedHat 7.6, gcc 9.2.0)
POWER: Summit-like POWER 9 machine (RedHat, gcc 9.3.0)
Average of 10 executions.
Fork-join benchmark: # of yields = 0
Yield benchmark: # of ULTs = 8
</p></details>

### Limitation

Scheduling a new ULT can need memory allocation, which can fail if the system does not have enough memory.  Currently Argobots does not handle this error.  This will be fixed in the future and at this point, this method should be enabled by default.

## Checklist

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -174,6 +174,12 @@ AC_ARG_ENABLE([aligned-alloc],
         [do not use aligned memory allocation]),,
         [enable_aligned_alloc=yes])
 
+# --enable-lazy-stack-alloc
+AC_ARG_ENABLE([lazy-stack-alloc],
+    AS_HELP_STRING([--enable-lazy-stack-alloc],
+        [lazily allocate and eagerly release ULT stacks]),,
+        [enable_lazy_stack_alloc=no])
+
 # --enable-feature
 AC_ARG_ENABLE([feature],
 [  --enable-feature@<:@=OPTS@:>@  enable/disable features.
@@ -618,6 +624,10 @@ AS_IF([test "x$enable_aligned_alloc" != "xno"],
     [AC_DEFINE(ABT_CONFIG_USE_ALIGNED_ALLOC, 1,
         [Define to allocate objects aligned to the cache line size])])
 
+# --enable-lazy-stack-alloc
+AS_IF([test "x$enable_lazy_stack_alloc" != "xyes"],
+    [AC_DEFINE(ABT_CONFIG_DISABLE_LAZY_STACK_ALLOC, 1,
+        [Define to disable lazy stack allocation])])
 
 # --enable-feature
 save_IFS="$IFS"
--- a/src/include/abt.h.in
+++ b/src/include/abt.h.in
@@ -631,6 +631,8 @@ enum ABT_info_query_kind {
     ABT_INFO_QUERY_KIND_ENABLED_STACK_OVERFLOW_CHECK,
     /** Wait policy */
     ABT_INFO_QUERY_KIND_WAIT_POLICY,
+    /** Whether a ULT stack is lazily allocated by default or not */
+    ABT_INFO_QUERY_KIND_ENABLED_LAZY_STACK_ALLOC,
 };
 
 /**
--- a/src/include/abtd_fcontext.h
+++ b/src/include/abtd_fcontext.h
@@ -78,10 +78,9 @@ static void ABTD_ythread_context_func_wrapper(fcontext_t *p_fctx)
 }
 
 static inline void ABTD_ythread_context_init(ABTD_ythread_context *p_ctx,
-                                             void *p_stack, size_t stacksize)
+                                             void *p_stacktop, size_t stacksize)
 {
     ABTDI_fcontext_init(&p_ctx->ctx);
-    void *p_stacktop = (void *)(((char *)p_stack) + stacksize);
     p_ctx->p_stacktop = p_stacktop;
     p_ctx->stacksize = stacksize;
     ABTD_atomic_relaxed_store_ythread_context_ptr(&p_ctx->p_link, NULL);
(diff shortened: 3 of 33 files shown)


shintaro-iwasaki:
test:argobots/all
test:argobots/freebsd
test:argobots/solaris
test:argobots/osx

shintaro-iwasaki:
test:argobots/all
test:argobots/freebsd
test:argobots/solaris
test:argobots/osx

shintaro-iwasaki:
test:argobots/all
test:argobots/freebsd
test:argobots/solaris
test:argobots/osx

Issue #361, opened 2021-08-10
mutexes busy spin by default even when using _WAIT scheduler and pool variants


carns:
See discussion on mailing list at https://lists.argobots.org/pipermail/discuss/2021-July/000157.html.

Confirmed with a Margo unit test in a development branch at https://github.com/mochi-hpc/mochi-margo/blob/carns/dev-scheduling-test/tests/unit-tests/margo-scheduling.c.  The following example shows swapping out two argobots libraries with the only change being whether the simple mutex implementation is enabled (default) or disabled.  This configuration change is enacted with a modified spack package that includes a "simple-mutex" variant. The test program deliberately sets up a scenario in which a ULT will attempt to lock a mutex that is held by another ULT for roughly 5 seconds.

The default behavior consumes user CPU time for the duration of the test case.  The simple mutex consumes negligible user CPU time.

```
> spack load -r argobots+simple-mutex
> tests/unit-tests/margo-scheduling
Running test suite with seed 0xdbb2a8d1...
/margo/abt_mutex_cpu                 user CPU time used: 5.086535
[ OK    ] [ 5.20039447 / 5.08659853 CPU ]
1 of 1 (100%) tests successful, 0 (0%) test skipped.
> spack  unload argobots
> spack load -r argobots~simple-mutex
> tests/unit-tests/margo-scheduling
Running test suite with seed 0x45d7f56c...
/margo/abt_mutex_cpu                 user CPU time used: 0.001290
[ OK    ] [ 5.20056152 / 0.00491515 CPU ]
1 of 1 (100%) tests successful, 0 (0%) test skipped.
```

This is problematic for I/O oriented use cases (Margo and more generally Mochi) in which ULTs are really just blocking on peripheral resource availability.  Sometimes our execution streams are over-provisioned according to the characteristics of the platform, and we would prefer to use CPU time for other purposes (or make it available for co-located processes). I don't have a specific application example for  which this is a problem, but we suspect that it has confused some debugging efforts in the past by causing some nominally "just waiting for access to an I/O device" execution streams to consume CPU.

I don't know what the implications are of the mutex implementation (beyond the CPU usage).  For our use case in isolation it would be better (in order of preference) to a) toggle the default behavior or b) have the capability to toggle it at runtime (so that we can gain this capability without altering the argobots build) or c) if it is a build-time option, make it accessible via upstream spack variant.

[...]

PR #363, opened 2021-08-15
configure: disable simple mutex by default

shintaro-iwasaki:
## Pull Request Description

This PR sets `--disable-simple-mutex` at configure time by default.

### History

Because of several bugs in "the non-simple mutex" (#14, #34, #41), the simple mutex became the default implementation for correctness (#102). The simple mutex is yield-based, so most users do not expect this behavior (see #361). The "non-simple mutex" has been reasonably fixed by #268, which can be enabled by setting "--disable-simple-mutex". As requested by several users, this PR now changes the default mutex implementation.

### Performance

The previous default mutex (= `--enable-simple-mutex`) should be slightly faster if the mutex is not contended. The new default mutex (= `--disable-simple-mutex`) should be more efficient if contention happens.  Since some synchronization objects such as `ABT_rwlock` internally uses the logic of `ABT_mutex`, this affects their performance.

If you like the original behavior, please set `--enable-simple-mutex` at configure time, though I believe the new default mutex should perform reasonably well in most cases.

## Checklist

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -220,10 +220,10 @@ AC_ARG_ENABLE([wait-policy],
         passive      - suspend CPU cores on blocking.  It increases latency.
 ],,[enable_wait_policy=auto])
 
-# --disable-simple-mutex
+# --enable-simple-mutex
 AC_ARG_ENABLE([simple-mutex],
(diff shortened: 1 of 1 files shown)


shintaro-iwasaki:
test:argobots/all

PR #364, opened 2021-08-16
eventual: provide a static initializer for eventual

### LINKED ISSUE #362: Suggestion: add an ABT_EVENTUAL_INITIALIZER and ABT_eventual_memory
If it's possible, adding the same mechanism as for mutex and condition variables to initialize an eventual in static memory would be very useful.

shintaro-iwasaki:
## Pull Request Description

This PR fixes #362. Specifically, this PR implements a static initializer for `ABT_eventual`. The static initializer is cheaper than `ABT_eventual_create()`. There's no need to explicitly free a statically initialized eventual, so it might be easy to handle in some cases.

Note that a statically initialized `ABT_eventual` does not accept `nbytes`. Setting a value is prohibited.

@mdorier: For possible future optimizations, `ABT_eventual` functions require Argobots initialization unlike `ABT_mutex` and `ABT_cond`; some (not all) `ABT_mutex/cond` functions can be called before `ABT_init()` and after `ABT_finialize()` (see #299). If you need this for `ABT_eventual`, please leave a comment.

## Checklist

### DIFF
--- a/src/eventual.c
+++ b/src/eventual.c
@@ -44,6 +44,9 @@ int ABT_eventual_create(int nbytes, ABT_eventual *neweventual)
     ABTI_UB_ASSERT(ABTI_initialized());
     ABTI_UB_ASSERT(neweventual);
 
+    /* Check if the size of ABT_eventual_memory is okay. */
+    ABTI_STATIC_ASSERT(sizeof(ABTI_eventual) <= sizeof(ABT_eventual_memory));
+
     int abt_errno;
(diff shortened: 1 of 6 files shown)


shintaro-iwasaki:
test:argobots/all

mdorier:
Thanks for this PR, I'm looking forward to using this feature. I don't think I need to manipulate eventuals outside of `ABT_init`/`finalize` scope.

Issue #355, opened 2021-07-26
Utilize the new pool interface


shintaro-iwasaki:
#354 introduced a new pool interface, which is more flexible and easy to implement.  We should implement the following to fully utilize this new pool interface.

- New user APIs (mainly pool routines) that accept `ABT_pool_context`
- Clarification and revision of API descriptions
- Predefined random work-stealing pool (`ABT_POOL_RANDWS`)
- Predefined priority-aware pool `ABT_POOL_PRIO_FIFO` (or maybe `ABT_POOL_PRIO_RANDWS` ...?)
- More examples (ring-buffer, LIFO, fine-grained priority control, ...)

shintaro-iwasaki:
`ABT_pool` functions that accept `ABT_pool_context` were implemented by #357.
A predefined random work-stealing pool (`ABT_POOL_RANDWS`) was implemented by #359.

shintaro-iwasaki:
Some users are still relying on the old customized pool API (see #366). We should create a unit test similar to https://github.com/pmodels/argobots/blob/main/test/basic/pool_user_def.c for the old custom pool API.

PR #368, opened 2021-10-15
pool: Add a unit test for the old pool API

epmikida:
This API is still used by some users according to issue so we want to make sure it is still tested.
See #355.


shintaro-iwasaki:
Note: I relaxed the GitHub Action restriction, so the basic CI checks should run automatically when you push/rebase commits to your local branch registered to this PR.

PR #371, opened 2021-11-11
adjust include path for 128-bit CAS autoconf test

### LINKED ISSUE #370: autoconf 128 CAS test fails unless configure is executed from within source tree
To reproduce, run configure (on a modern system) the following two ways and compare the contents of `src/include/abt_config.h`:

```
carns-x1-7g ~/w/s/mochi> cd argobots
carns-x1-7g ~/w/s/m/argobots (main=)> mkdir build
carns-x1-7g ~/w/s/m/argobots (main=)> cd build
carns-x1-7g ~/w/s/m/a/build (main=)> ../configure > /dev/null
carns-x1-7g ~/w/s/m/a/build (main=)> cd ../
carns-x1-7g ~/w/s/m/argobots (main=)> ./configure > /dev/null
carns-x1-7g ~/w/s/m/argobots (main=)> 
diff src/include/abt_config.h build/src/include/abt_config.h 
69c69
< #define ABT_CONFIG_HAVE_ATOMIC_INT128 1
---
> /* #undef ABT_CONFIG_HAVE_ATOMIC_INT128 */
```
The problem is the presence of a relative (to the source tree) include path in the autoconf test at https://github.com/pmodels/argobots/blob/main/configure.ac#L1111.

The "out-of-tree" configure/build style is the more prevalent approach in many packaging environments (I assume including Spack, though I haven't confirmed yet).  This is possibly affecting code paths like that described in #369 .

I just wanted to open an issue to document it for the moment; I think I can contribute a PR to fix it in the next couple of days and report if it changes the behavior of #369 or not.

carns:
This PR modifies the CFLAGS used in configure.ac temporarily for the 128-bit CAS test.  This is necessary because the test relies on a header file within the source directory.  Without this PR the 128-bit CAS test may erroneously fail for out-of-tree builds.

Fixes #370

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -1108,6 +1108,11 @@ AS_IF([test x"$use_huge_page" = x"yes"],
 # check for the availability of int128 CAS. It is hard to get availability of
 # instructions (and compiler support), so we rely on AC_RUN_IFELSE().
 atomic_int128_cas_mode="disabled"
+# Note: temporarily prepend the source directory to the include path so that
+# the following configure test can locate abtd_asm_int128_cas.h regardless
+# of where the configure script is executed.
+SAVE_CFLAGS="$CFLAGS"
+CFLAGS="-I$srcdir $CFLAGS"
 AC_RUN_IFELSE([AC_LANG_PROGRAM([#include "src/include/asm/abtd_asm_int128_cas.h"],[
 __int128 val, old, neu;
 int ret1, ret2;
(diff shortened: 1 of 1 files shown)


shintaro-iwasaki:
test:argobots/all
test:argobots/osx
test:argobots/freebsd
test:argobots/solaris

shintaro-iwasaki (approved):
Yes the problem is fixed as far as I checked on my laptop. Thank you for your contribution!

